PR #6, opened 2018-12-14
Kill the native folder

shaundon:


ios/native.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
matthewdavidson: should this be committed?

PR #11, opened 2018-12-18
Bpk 1754 more rn upgrade

shaundon:
* Cleaned up storybook logic a lot.
* Made RN packager handle the storybook, no more calling it directly.
* Side effect of this is that we no longer get the web interface at localhost:7007. **We need to use the device itself to control things. This is going to be slightly slower for taking screenshots en masse but I think that's an acceptable trade off for now.** We can always try and get it working again in future.
* Had to add a load of polyfills to Android. I'm not sure what's changed to make this necessary. My only guess is that Storybook 4 is doing stuff that Android's JS runtime can't handle. But then why would that be in there? The mind baffles, it really does. Apparently Storybook 5, due in January, might fix this. YAY

**Still to do**
* Get tests working without breaking storybook
* Port web-based addons to run on device (I think this can be done in a future PR. We can live without these for the moment IMO).


index.android.js
```diff
@@ -17,7 +17,16 @@
  */
 
 import { AppRegistry } from 'react-native';
+import Storybook from './storybook';
 
-import App from './app/index';
+// https://github.com/facebook/react-native/issues/15902
+// Symbol polyfills
+global.Symbol = require('core-js/es6/symbol');
```
matthewdavidson: I'm reading the docs for core-js usage because I was confused between the `core-js/es6/..` imports vs `core-js/fn/..` imports.

Can't you just use `import 'core-js/es6/symbol';` and that should be enough? same with map, set and array find?

shaundon: I'll give that a go

shaundon: Yep, did the trick!

PR #12, opened 2018-12-19
[BPK-1754] Update screenshots

shaundon:
This took a long time.


packages/react-native-bpk-component-navigation-bar/src/isIphoneX.js
```diff
@@ -17,9 +17,13 @@
...
 /* @flow */
 
-import { DeviceInfo } from 'react-native';
+import { DeviceInfo, Platform } from 'react-native';
 
 export default (() => {
+  if (Platform.OS === 'android') {
+    return false;
+  }
+
```
shaundon: Importing this in stories was crashing Android so this seemed like the easiest fix. An alternative approach would be to rename this file `isIphoneX.ios.js` and then having a `.android.js` one that just returned false, but that feels like overengineering.

packages/react-native-bpk-component-navigation-bar/stories.js
```diff
@@ -35,6 +37,20 @@ import BpkNavigationBar, {
...
+/*
+HACK
+On iOS, Storybook's UI covers the status bar which affects the appearance
+of the navigation bar. This pulls it upwards to counteract it.
+*/
+const IOS_STATUS_BAR_HEIGHT = isIphoneX ? 44 : 20;
+const styles = StyleSheet.create({
+  navigationBar: Platform.select({
+    ios: {
+      marginTop: -IOS_STATUS_BAR_HEIGHT,
```
shaundon: Screenshots in this PR don't show iPhone X, so here's proof it works:

simulator screen shot - iphone xs - 2018-12-20 at 09 43 34

PR #21, opened 2019-01-03
Update fonts.gradle to correct bpk-svg default

micetti:
When installing the backpack icons with npm and following the installation steps for android the icons won't be installed because of the path to bpk-svgs which gets installed in the node_modules. The need for using gradle options to make this work is not mentioned in the installation steps

### DIFF
--- a/packages/react-native-bpk-component-icon/fonts.gradle
+++ b/packages/react-native-bpk-component-icon/fonts.gradle
@@ -4,7 +4,7 @@
 
 def config = project.hasProperty("bpkicon") ? project.bpkicon : [];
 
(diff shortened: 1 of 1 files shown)


tiagoengel:
Sorry @micetti I don't really understand the reasoning behind this. `node_modules` should normally be at the root level, hence `./node_modules/`. Can you explain the reasoning behind this change?

tiagoengel:
Got it @micetti. Unfortunately I'll have to close this as we can't change the default because this is how it's used internally by our apps. 

You can use the `iconFontsDir` to change it in your app.

PR #22, opened 2019-01-03
[BPK-2136] Make message optional for BpkBadge

shaundon:
We didn't anticipate that users would create badges with **only** icons, no text. This causes two issues:

1. A proptypes warning, as `message` is required.
2. Weird spacing, as each of the icons in the accessory view has `marginEnd: spacingSm`. This looks fine with text but isn't required when there is no text.

This PR does the following:

1. Makes `message` optional.
2. Adds an `accessibilityLabel` prop, which is required if `message === null`.
3. Adds an example of badges with **only** icons to the storybook, tests and associated screenshots.
4. Changes the way margin is applied. The last icon in the accessory view now gets no `marginEnd`, and we instead apply `marginStart` to the text next to it instead. You can see that it works from the screenshots included in this PR.


matthewdavidson:
Do icons or badges have an `accessibilityLabel`? Having no text inside them seems like an a11y nightmare

georgegillams:
LGTM, but we need to revisit `accessibilityLabel`.

@matthewdavidson we did spend some time discussing that. Not sure what the best approach is. One option would be to use custom prop types to ensure that either `accessibilityLabel` or `message` is set? Bit like we do for Buttons?

georgegillams:
Also the build is failing with flow errors 😢

shaundon:
> Do icons or badges have an accessibilityLabel? Having no text inside them seems like an a11y nightmare

@matthewdavidson so yeah as @georgegillams said, we chatted about this. Ideal situation would be to add a required `accessibilityLabel` prop but I was hesitant to make a breaking change over it, which that would have to be. 

So I looked at badges for web and they don't have any a11y considerations either, and then we were thinking that badges are more of a visual accessory so maybe forcing accessibility labels on them isn't desirable and it's ok to do it this way.

Having said that, if the consensus is that we should make it required and just do the breaking change anyway, happy to go ahead, and we can put in a PR to consumers of it to minimise the pain.

shaundon:
Also worth mentioning, we can make `accessibilityLabel` fall back to using `message` as the default (this is how buttons do it). So it would only be a required prop if the user hasn't passed in a message. That way it won't be breaking for most consumers, as they'll be using `message`.

shaundon:
> Also the build is failing with flow errors 😢

Now fixed

matthewdavidson:
> Also worth mentioning, we can make accessibilityLabel fall back to using message as the default (this is how buttons do it). So it would only be a required prop if the user hasn't passed in a message. That way it won't be breaking for most consumers, as they'll be using message.

I like this.

packages/react-native-bpk-component-badge/stories.js
```diff
@@ -96,21 +96,38 @@ const generateBadgeStory = (
...
+                  />
+                ) : null
+              }
+              docked={config.docked}
+              type={i}
+              style={style.badge}
+            />
+            {config.icons === 'multiple' && (
+              <BpkBadge
+                // accessibilityLabel={content}
```
matthewdavidson: Commented out?

packages/react-native-bpk-component-badge/src/BpkBadgeIcons.js
```diff
@@ -47,10 +50,10 @@ const BpkBadgeIcons = (props: Props) => {
     itemStyle.push(StyleSheet.flatten(userItemStyle));
   }
 
-  const adjustedIcons = icons.map(icon =>
+  const adjustedIcons = icons.map((icon, index) =>
     React.cloneElement(icon, {
       small: true,
-      style: itemStyle,
+      style: [itemStyle, index === icons.length - 1 ? styles.lastItem : null],
```
matthewdavidson: Where are your css pseudo selectors when you need em? :P

packages/react-native-bpk-component-badge/stories.js
shaundon: Good catch! I was testing that the prop types warning came up. Pushing an update now.

packages/react-native-bpk-component-badge/src/BpkBadgeIcons.js
shaundon: I know right!

PR #25, opened 2019-01-09
[BPK-2126] Calendar bridge

georgegillams:
# Roadmap

## iOS
 - [x] Create a boilerplate component for each platform
 - [x] Create iOS source files/project for bridging
 - [x] ~Add BPKCalendar cocoa dependency to iOS source files/project~
 - [x] Define setup / build processes etc for `react-native-bpk-component-calendar` given it's use of iOS subproject
 - [x] Implement react native iOS bridge component

## Android
 - [ ] Create Android source files/project for bridging
 - [ ] Implement react native Android bridge component
 - [ ] Try to combine RN iOS and Android bridge components (should have the same API)

# Current questions
 - [x] How should our `.gitignore` file look by the end of this?
 - [x] ~How do we ensure that the latest version of BPKCalendar is being referenced by the iOS bridging project?~
 - [x] Can we combine the RN iOS and Android bridge components (should have the same API)?

# Todo (added by Shaun)

- [x] Add lots more tests
- [x] Custom prop type for `selectedDates`
- [x] enum for selection type
- [x] Readme
- [x] Add Flow types
- [x] Common types file to share with Android
- [x] Add lots more stories including ones for screenshots
- [x] Replace hard-coded 300 and 500 from iOS side
- [x] Make locale required on user side
- [x] Determine if we should be setting defaults for minDate and maxDate or requiring them from the user and change if necessary
- [x] Protect against mindate that's greater than maxdate (app crashes if we try and do that)
- [ ] Document possible locales
- [x] Fix version number
- [x] Add missing dev deps
- [x] Rename props to fit with conventions
- [x] Fix issue where setting `selectedDates` has no effect on the calendar (it doesn't highlight them). You can see this by setting `selectedDates` to an initial value and seeing that the day doesn't get highlighed.
- [x] When we switch selectionType the calendar looks weird until you click about. Maybe we need to force a re-render.
- [x] Export the selectionType enum over the bridge so it isn't duplicated on the iOS and JS side.
- [x] Rename Obj-c files - `RCT` prefix is for internal React code and we shouldn't use it.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1,4 +1,6 @@
 PODS:
+  - Backpack (6.0.0):
+    - FSCalendar (~> 2.8)
   - boost-for-react-native (1.63.0)
   - BVLinearGradient (2.4.0):
     - React
@@ -7,9 +9,13 @@ PODS:
     - boost-for-react-native
     - DoubleConversion
     - glog
+  - FSCalendar (2.8.0)
   - glog (0.3.5)
   - React (0.57.7):
     - React/Core (= 0.57.7)
+  - react-native-bpk-component-calendar (0.0.1):
+    - Backpack (~> 6.0.0)
+    - React
   - react-native-maps (0.23.0):
     - React
   - React/Core (0.57.7):
@@ -50,10 +56,12 @@ PODS:
   - yoga (0.57.7.React)
 
 DEPENDENCIES:
+  - Backpack (~> 6.0.0)
   - BVLinearGradient (from `../node_modules/react-native-linear-gradient`)
   - DoubleConversion (from `../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec`)
   - Folly (from `../node_modules/react-native/third-party-podspecs/Folly.podspec`)
   - glog (from `../node_modules/react-native/third-party-podspecs/glog.podspec`)
+  - react-native-bpk-component-calendar (from `../packages/react-native-bpk-component-calendar`)
   - react-native-maps (from `../packages/react-native-bpk-component-map/node_modules/react-native-maps`)
   - React/Core (from `../node_modules/react-native`)
   - React/CxxBridge (from `../node_modules/react-native`)
@@ -67,7 +75,9 @@ DEPENDENCIES:
 
 SPEC REPOS:
   https://github.com/cocoapods/specs.git:
+    - Backpack
     - boost-for-react-native
+    - FSCalendar
 
 EXTERNAL SOURCES:
   BVLinearGradient:
@@ -80,21 +90,26 @@ EXTERNAL SOURCES:
     :podspec: "../node_modules/react-native/third-party-podspecs/glog.podspec"
   React:
     :path: "../node_modules/react-native"
+  react-native-bpk-component-calendar:
+    :path: "../packages/react-native-bpk-component-calendar"
   react-native-maps:
     :path: "../packages/react-native-bpk-component-map/node_modules/react-native-maps"
   yoga:
     :path: "../node_modules/react-native/ReactCommon/yoga"
 
 SPEC CHECKSUMS:
+  Backpack: a9a3c704beed3e180c19ee1d1c26eeae2713f49f
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   BVLinearGradient: 6d8909e3fc6b089defa4b89d967441fa6827a2ee
   DoubleConversion: bb338842f62ab1d708ceb63ec3d999f0f3d98ecd
   Folly: c89ac2d5c6ab169cd7397ef27485c44f35f742c7
+  FSCalendar: 3a5ed3636e2ba1b83ebebfcf0c2603105a6f5efe
   glog: e8acf0ebbf99759d3ff18c86c292a5898282dcde
   React: 1fe0eb13d90b625d94c3b117c274dcfd2e760e11
+  react-native-bpk-component-calendar: 97dc3fe80e416867b37a7109c379a3defb564af6
   react-native-maps: 066c2afcc89e18726377bcc685315f989ca22449
   yoga: b1ce48b6cf950b98deae82838f5173ea7cf89e85
 
-PODFILE CHECKSUM: a820fd1dbae2729bbf89b107411a41e83d79792a
+PODFILE CHECKSUM: 38726a947f3ee65935c1fb42800e2e5243fb0d0d
 
 COCOAPODS: 1.5.3
--- a/packages/react-native-bpk-component-calendar/README.md
+++ b/packages/react-native-bpk-component-calendar/README.md
@@ -1,0 +1,75 @@
+# react-native-bpk-component-calendar
+
+> Backpack React Native calendar component.
+
+## Installation
+
+```sh
+npm install react-native-bpk-component-calendar --save-dev
+```
+
+Because this package ships with native code, it is also necessary to add some native dependencies to your RN project:
+
+### Android
+TODO - NEED TO WRITE THIS BIT
+
+### iOS
+Having installed the NPM package, you need to add the following dependencies to your Podfile:
+
+ - `react-native-bpk-component-calendar`, using a relative path via `node_modules`
+
+For example:
+```
+  pod 'react-native-bpk-component-calendar', path: '../node_modules/react-native-bpk-component-calendar'
+```
+
+## Usage
+
+```js
+import React, { Component } from 'react';
+import { StyleSheet, View } from 'react-native';
+import { spacingBase } from 'bpk-tokens/tokens/base.react.native';
+import BpkCalendar, { SELECTION_TYPES } from 'react-native-bpk-component-calendar';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    padding: spacingBase,
+  },
+});
+
+export default () => (
+  <View style={styles.container}>
+    <BpkCalendar
+      selectionType={SELECTION_TYPES.single}
+      selectedDates={[new Date()]}
+      onChangeSelectedDates={newDates => {
+        console.warn("Date selection changed");
+      }}
+    />
+  </View>
+);
+```
+
+## Props
+
+### BpkCalendar
+
+| Property                | PropType               | Required   | Default Value          |
+| ----------------------- | ---------------------- | ---------- | ---------------------- |
+| locale                  | string                 | true       | -                      |
+| maxDate                 | Date                   | false      | null                   |
+| minDate                 | Date                   | false      | null                   |
+| onChangeSelectedDates   | function               | false      | null                   |
+| selectedDates           | Array(Date)            | false      | null                   |
+| selectionType           | oneOf(SELECTION_TYPES) | false      | SELECTION_TYPES.single |
+
+#### locale
+
+TODO explain locale
+
+#### selectedDates
+
+* When `selectionType` is `SELECTION_TYPES.single`, you should only include zero or one entries in the `selectedDates` array.
+* When `selectionType` is `SELECTION_TYPES.range`, you should only include zero, one or two entries in the `selectedDates` array.
--- a/packages/react-native-bpk-component-calendar/package.json
+++ b/packages/react-native-bpk-component-calendar/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "react-native-bpk-component-calendar",
+  "private": true,
+  "version": "0.0.1",
+  "main": "index.js",
+  "description": "Backpack React Native calendar component.",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^16.0.0-alpha.1",
+    "react-native": ">= 0.47.0"
+  },
+  "dependencies": {
+    "bpk-tokens": "^27.4.4",
+    "prop-types": "^15.5.8"
+  },
+  "devDependencies": {
+    "react-native-bpk-component-picker": "^3.0.16",
+    "react-native-bpk-component-select": "^2.1.13",
+    "react-native-bpk-component-text-input": "^3.0.11"
+  }
+}
--- a/packages/react-native-bpk-component-calendar/react-native-bpk-component-calendar.podspec
+++ b/packages/react-native-bpk-component-calendar/react-native-bpk-component-calendar.podspec
@@ -1,0 +1,20 @@
+require 'json'
+
+package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
+
+Pod::Spec.new do |s|
+  s.name         = "react-native-bpk-component-calendar"
+  s.version      = package['version']
+  s.summary      = "Backpack calendar component for React Native"
+
+  s.authors      = { "backpack" => "backpacksquad@skyscanner.net" }
+  s.homepage     = "https://backpack.github.io/components/calendar?platform=ios"
+  s.license      = "MIT"
+  s.platform     = :ios, "10.0"
+
+  s.source       = { :git => "https://github.com/Skyscanner/backpack-react-native.git" }
+  s.source_files  = "src/ios/CalendarBridge/**/*.{h,m}"
+
+  s.dependency 'React'
+  s.dependency 'Backpack', '~> 6.0.0'
+end
--- a/packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js
+++ b/packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js
@@ -1,0 +1,216 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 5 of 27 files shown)


packages/react-native-bpk-component-calendar/react-native-bpk-component-calendar.podspec
```diff
@@ -0,0 +1,19 @@
...
+
+  s.authors      = { "backpack" => "backpacksquad@skyscanner.net" }
+  s.homepage     = "https://backpack.github.io/components/calendar"
+  s.license      = "MIT"
+  s.platform     = :ios, "9.0"
+
+  s.source       = { :git => "https://github.com/Skyscanner/backpack-react-native.git" }
+  s.source_files  = "src/ios/CalendarBridge/**/*.{h,m}"
+
+  s.dependency 'React'
```
georgegillams: Do we need this dep here. By definition, the RN codebase to which this library is added has React already.

packages/react-native-bpk-component-calendar/react-native-bpk-component-calendar.podspec
```diff
@@ -0,0 +1,19 @@
...
+
+Pod::Spec.new do |s|
+  s.name         = "react-native-bpk-component-calendar"
+  s.version      = package['version']
+  s.summary      = "Backpack calendar component for React Native"
+
+  s.authors      = { "backpack" => "backpacksquad@skyscanner.net" }
+  s.homepage     = "https://backpack.github.io/components/calendar"
+  s.license      = "MIT"
+  s.platform     = :ios, "9.0"
```
georgegillams: Wasn't sure which platform to select, so I went with the one used in our example app podspec

packages/react-native-bpk-component-calendar/README.md
```diff
@@ -0,0 +1,68 @@
...
+    <BpkCalendar
+      minDate={new Date()}
+      maxDate={new Date()}
+      selectionType="single"
+      selectedDates={[new Date()]}
+      onDatesSelected={newDates => {
+        console.warn("Date selection changed");
+      }}
+    />
+  </View>
```
shaundon: We should make this example more useful, e.g. make `maxDate` a year from today and so on. (Or just remove the `maxDate` usage because it defaults to that)

packages/react-native-bpk-component-calendar/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "react-native-bpk-component-calendar",
+  "private": true,
+  "version": "0.0.17",
```
shaundon: Just out of interest where did `0.0.17` come from? Seems like a weird number for an unreleased component to have.

packages/react-native-bpk-component-calendar/react-native-bpk-component-calendar.podspec
```diff
@@ -0,0 +1,19 @@
...
+
+package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
+
+Pod::Spec.new do |s|
+  s.name         = "react-native-bpk-component-calendar"
+  s.version      = package['version']
+  s.summary      = "Backpack calendar component for React Native"
+
+  s.authors      = { "backpack" => "backpacksquad@skyscanner.net" }
+  s.homepage     = "https://backpack.github.io/components/calendar"
```
shaundon:
```suggestion
  s.homepage     = "https://backpack.github.io/components/calendar?platform=ios"
```

🙌

packages/react-native-bpk-component-calendar/react-native-bpk-component-calendar.podspec
shaundon: Let's just follow the lead of what other components that include bridges do, I guess.

packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js
```diff
@@ -0,0 +1,58 @@
...
+            testID="123" // <- Arbitrary prop.
+          />,
+        )
+        .toJSON();
+
+      expect(tree).toMatchSnapshot();
+    });
+  });
+};
+export default commonTests;
```
shaundon: I guess we should add more tests too, once the API is decided upon and stable.

packages/react-native-bpk-component-calendar/src/BpkCalendar.android.js
```diff
@@ -0,0 +1,67 @@
...
+
+  const style = [styles.base];
+  if (userStyle) {
+    style.push(userStyle);
+  }
+
+  /* eslint-disable */
+  return (
+    <View style={style} {...rest}>
+      <Text>THIS COMPONENT HAS NOT BEEN BRIDGED YET</Text>
```
shaundon: This is fine for now as we aren't going to release it in this state, but if we ever *do* release something only for one platform, we should probably do a nicer error message.

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
```diff
@@ -0,0 +1,95 @@
...
+export type Props = {
+  style: ?StyleObj,
+};
+
+const getDateInOneYear = () => {
+  const date = new Date();
+  const newMonth = date.getMonth() + 12;
+  date.setMonth(newMonth);
+  return date;
+};
```
shaundon: The naming of and within this function feels a bit generic, I think it could be more descriptive. What about something like:

```
const getDateOneYearFromNow = () => {
  const today = new Date();
  const monthOneYearFromNow = date.getMonth() + 12;
  return today.setMonth(monthOneYearFromNow);
};
```

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
```diff
@@ -0,0 +1,95 @@
...
+    </View>
+  );
+};
+
+BpkCalendar.propTypes = {
+  style: ViewPropTypes.style,
+  minDate: PropTypes.instanceOf(Date),
+  maxDate: PropTypes.instanceOf(Date),
+  selectedDates: PropTypes.arrayOf(PropTypes.instanceOf(Date)),
+  onDateSelectionChanged: PropTypes.func,
```
shaundon: I had a look at some other RN components and I think the convention is present tense for event-driven props, e.g. `onChangeText`, `onPress`, `onClose`.

How about naming it `onSelectionChange`? (I figure the `Date` bit is maybe redundant but feel free to disagree)

shaundon: These should be alphabetised.

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
```diff
@@ -0,0 +1,95 @@
...
+
+const styles = StyleSheet.create({
+  base: {
+    width: '100%',
+    height: '100%',
+  },
+});
+
+export type Props = {
+  style: ?StyleObj,
```
shaundon: Flow type doesn't match the prop types, probably just forgot to add them?

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const style = [styles.base];
+  if (userStyle) {
+    style.push(userStyle);
+  }
+
+  return (
+    <View style={style} {...rest}>
+      <RCTCalendarView
+        locale="en-GB"
```
shaundon: We probably can't hard code the locale here. We can make it a required prop instead and link users to a list of possible locales.

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
```diff
@@ -0,0 +1,95 @@
...
+        onDateSelection={event => {
```
shaundon: I _think_ it's more efficient to define this outside of the render block so that the function doesn't get recreated every time the component re-renders. @tiagoengel knows more about this than me though.

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
```diff
@@ -0,0 +1,95 @@
...
+          const datesConverted = event.nativeEvent.selectedDates.map(
+            dateString => new Date(Date.parse(dateString)),
```
shaundon: We might need to try/catch this or something for safety.

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
```diff
@@ -0,0 +1,95 @@
...
+  minDate: PropTypes.instanceOf(Date),
+  maxDate: PropTypes.instanceOf(Date),
+  selectedDates: PropTypes.arrayOf(PropTypes.instanceOf(Date)),
+  onDateSelectionChanged: PropTypes.func,
+};
+
+BpkCalendar.defaultProps = {
+  style: null,
+  minDate: new Date(),
+  maxDate: getDateInOneYear(),
```
shaundon: Can we make these assumptions? Is this what the underlying iOS class does too? We could always force users to supply min and max dates themselves.

packages/react-native-bpk-component-calendar/src/__snapshots__/BpkCalendar-test.ios.js.snap
```diff
@@ -0,0 +1,41 @@
...
+        "flex": 1,
+      },
+    ]
+  }
+  testID="123"
+/>
+`;
+
+exports[`iOS BpkCalendar should render correctly with custom style 1`] = `
+<View
```
shaundon: This looks a bit empty. Or is this just how Jest renders native code? I guess it would make sense that it doesn't render anything past the React code, because what would it put there..

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTCalendarViewManager.m
```diff
@@ -0,0 +1,71 @@
...
+RCT_EXPORT_MODULE()
+
+- (UIView *)view
+{
+    RCTCalendarView *calendar = [RCTCalendarView new];
+    
+    calendar.translatesAutoresizingMaskIntoConstraints = NO;
+    [NSLayoutConstraint activateConstraints:@[
+                                              [calendar.widthAnchor constraintEqualToConstant:300],
+                                              [calendar.heightAnchor constraintEqualToConstant:500]
```
shaundon: Where do we get 300 and 500 from?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTCalendarViewManager.m
```diff
@@ -0,0 +1,71 @@
...
+ * `RCTCalendarViewManager` acts as the delegate of all of the `RCTCalendarView` views. This is just one
+ * pattern and it's perfectly fine to call `onDateSelection` from the `RCTCalendarView` directly.
+ */
+- (void)calendar:(RCTCalendarView *)calendar didChangeDateSelection:(NSArray<NSDate *> *)dateList
+{
+    if (!calendar.onDateSelection) {
+        return;
+    }
+    
+    NSDateFormatter *dateFormat = [[NSDateFormatter alloc] init];
```
shaundon: I reckon a comment here to explain why it's necessary and what's going on would be handy.

packages/react-native-bpk-component-calendar/package.json
georgegillams: It comes from the boilerplate 🤷‍♂️ 
https://github.com/Skyscanner/backpack-react-native/blob/master/packages/react-native-bpk-component-boilerplate/package.json#L4

packages/react-native-bpk-component-calendar/react-native-bpk-component-calendar.podspec
georgegillams: Good point!

packages/react-native-bpk-component-calendar/package.json
shaundon: Done

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
shaundon: Done

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
shaundon: Now alphabetised.

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
shaundon: Have made it a required prop. We can only get a list from the underlying OS, which may be a problem. Let's deal with it in a bit.

packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js
tiagoengel: The problem with this is that `onDateSelection` is always changing, so even if nothing really changed this will cause the react to interpret this as a new prop and re render the three. This is normally not a big unless it is part of a big three and it causes it to re-render. 

Not sure how much this affects things here as it is going directly to the 'bridged' component, plus the way things are now it's not possible to not do it. 

In order to avoid creating a new function every time we would have to change change `onDateSelection` accept `onDateSelectionChanged ` and return a "memoized" function that only receives the event. Something like this:

```
const onDateSelection = memoize(callback => {
    return (event) => {
       // do stuff
     }
}```

So for the same `onDateSelectionChanged` callback we would always have the same function.

But again this might not be required here so I wouldn't hold this PR because of this.

**Comment by tiagoengel** [comment]
Not sure I agree with the suggestion to add more components for the different selection types. As a consumer of this I would find the api easier the way it is now.

Having different components/HOCs that change the api based on the selection type would make it harder to use it in a situation where the user can change it.

For example, imagine there's a checkbox where you can choose the selection types and the calendar changes based on that. With the current API all that is needed is to provide the correct selection type prop. If we go for the other API consumers would have to implement themselves the logic to render different components based on the selection type.

**Comment by shaundon** [comment]
@tiagoengel fair enough - I'm happy to go with the majority here.

One tweak I would like to push for in this case is to go from three selection types - `single`, `multiple` and `range` down to just `multiple` and `range`. 

Because `selectedDates` is an array, we can infer `single` or `multiple` from `selectedDates.length` and then apply the selection type internally so that users don't need to worry about it.

How does that sound?

**Comment by tiagoengel** [comment]
You mean doing at the "js level"? Because the UI still needs to know if it is single or multiple to select more than one or change the selection when the user clicks on dates.

Not sure how that would work 🤔

**Comment by tiagoengel** [comment]
Hmm, my biggest question is still how the UI will know it should allow the user to select just one or multiple dates. E.g. if you start the calendar like this, (which I think is the most common situation)

```
<Calendar
    selectionType="single"
    selectionDates={[]} />
```

And you want it to allow only one date to be selected, how would you do it with only "multiple" and "range"?

**Comment by shaundon** [comment]
Sorry, I think I misunderstood what was happening behind the scenes before, I get it now.

We need the user to specify single or multiple because that'll determine how many dates are returned in the `onDateSelectionChanged` callback.

I'll keep them all separate 👍

File: packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js (hunk 6 above)
Comment (shaundon) [reply packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js]: Added more tests

File: packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js (hunk 13 above)
Comment (shaundon) [reply packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js]: Added an `if` block.

File: packages/react-native-bpk-component-calendar/src/__snapshots__/BpkCalendar-test.ios.js.snap (hunk 15 above)
Comment (shaundon) [reply packages/react-native-bpk-component-calendar/src/__snapshots__/BpkCalendar-test.ios.js.snap]: Fixed

File: packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js (hunk 14 above)
Comment (georgegillams) [reply packages/react-native-bpk-component-calendar/src/BpkCalendar.ios.js]: I don't think what the underlying component does is entirely relevant here. I think we should do what makes the most sense to our RN consumers

File: ios/Podfile.lock [hunk 18]
```diff
@@ -7,11 +9,14 @@ PODS:
     - boost-for-react-native
     - DoubleConversion
     - glog
+  - FSCalendar (2.8.0)
   - glog (0.3.5)
   - React (0.57.7):
     - React/Core (= 0.57.7)
   - react-native-maps (0.23.0):
```
georgegillams: This bit
```
  - react-native-maps (0.23.0):
    - React
```

Should be below
```
   - react-native-bpk-component-calendar (0.0.1):
     - React
```
instead of above.

That's all that's wrong with this podfile 😉

packages/react-native-bpk-component-calendar/stories.js
```diff
@@ -0,0 +1,198 @@
...
+    this.setState({
+      selectedDates: newDates,
+    });
+  };
+
+  render() {
+    const { range } = this.props;
+    const { selectedDates } = this.state;
+    return (
+      <View>
```
georgegillams:
```suggestion
      <View style={{ flex: 1 }} >
```

k0nserv: Is this needed?

georgegillams: screenshot 2019-01-23 at 14 37 27

k0nserv: Hmm I thought I had resolved that. Good catch

tiagoengel: Btw, I've improved (or so I think) this a bit in my branch so I'll add that as soon as this gets merged

PR #29, opened 2019-01-17
[NOJIRA] Update avd instructions for Android 9

matthewdavidson:


android/app/src/main/AndroidManifest.xml
```diff
@@ -13,6 +13,7 @@
         android:label="@string/app_name"
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true"
```
matthewdavidson: See https://stackoverflow.com/questions/45940861/android-8-cleartext-http-traffic-not-permitted/50834600#50834600

PR #31, opened 2019-01-18
Bpk 2188 fix android ripple

matthewdavidson:


packages/react-native-bpk-component-card/src/BpkCard.android.js
```diff
@@ -79,7 +79,11 @@ const BpkCard = props => {
 
   return (
     <View style={style}>
-      <BpkTouchableNativeFeedback accessibilityComponentType="button" {...rest}>
+      <BpkTouchableNativeFeedback
+        borderlessBackground={Platform.Version !== 28}
```
georgegillams: Should this condition be checked by `BpkTouchableNativeFeedback` itself so that we don't have to put it on every use?

matthewdavidson: Because some components desire the borderless background behaviour regardless of api version, BpkButtonLink for example:

button-link

matthewdavidson: You could ask - why not just set it to `false`.

I was trying to limit the blast radius of this fix but perhaps when api level 29 comes out it will surface again 🤔

georgegillams:
I've tested this out locally and it all looks 💯. Just got 1 small query above...

packages/react-native-bpk-component-card/src/BpkCard.android.js
georgegillams: Oh ok cool. tbh I assumed that wouldn't be desirable behaviour for any components! My bad!

PR #33, opened 2019-01-21
[BPK-2127] Upgrate to androidx

tiagoengel:


android/app/build.gradle
```diff
@@ -169,27 +169,28 @@ android {
...
     // From node_modules
     implementation 'com.facebook.react:react-native:+'
     // From node_modules
-    implementation 'net.skyscanner.backpack:Backpack:0.2.0'
-    implementation project(':react-native-maps')
+    implementation 'com.github.skyscanner:backpack-android:2.3.0'
+    // TODO: Normally this should use project(:react-native-maps) but it does not work with androidx
+    // version is 0.23. In any case this is what the app's code base is using, so it is a good idea
+    // we do the same
+    implementation 'com.airbnb.android:react-native-maps:0.17.1'
```
tiagoengel: This is how things are working in the app so it's probably better we keep the same here too. At least until `react-native-maps` upgrades to be compatible with androidX. 

I will have to check this better but I believe they only publish the java code when the bridge code changes, that is why using `0.17.1` here and `0.22.0` for the js code is fine. But will investigate this more... In any case this works fine

PR #35, opened 2019-01-22
[BPK-2183] Align Text examples between platforms

georgegillams:
To unfold all diffs:
```
const toClick = document.getElementsByClassName("btn btn-sm BtnGroup-item tooltipped tooltipped-n rendered js-rendered");
for (let i = toClick.length - 1; i >= 0; i -= 1){
toClick[i].click();
}
```


packages/react-native-bpk-component-text/stories.js
```diff
@@ -33,91 +33,93 @@ import CenterDecorator from '../../storybook/CenterDecorator';
 
 import BpkText, { WEIGHT_STYLES } from './index';
 
+const text = 'Lorem ipsum';
```
k0nserv: This should be in UPPER_SNAKE_CASE to indicate the constant nature of this value.

PR #37, opened 2019-01-28
Publish android native modules

tiagoengel:


CONTRIBUTING.md
```diff
@@ -228,6 +228,37 @@ Be aware that if `bpk-tokens` has changed, *all* packages in the repository will
 
 When a component is released for the first time on npm, remember to add the component to the Skyscanner organisation through the [npm UI](https://www.npmjs.com/settings/skyscanner/teams/team/backpack-react-native/access).
 
+### Native Android bridges
+
+Android bridges should be compiled and published into the internal archifatory, to do so after the js package has been released, run:
+
+```
+cd android
```
georgegillams: Can we not use `npm postpublish` to do this automatically every time `react-native-bpk-component-calendar` is published. Would obviously require a prompt to enter credentials if not already specified

tiagoengel: We can automate yes, but I'd rather do that in a different PR/card

android/gradle-maven-push.gradle
```diff
@@ -0,0 +1,62 @@
...
+                artifactId project.name
+                version version
+                description description
+
+                artifact bundleReleaseAar
+                artifact sourceJar
+
+                pom.withXml {
+
+                    def dependenciesNode = asNode().appendNode('dependencies')
```
amolgupta: this can be moved to a function.

tiagoengel: It can, but I don't see the benefit, I don't think we are going to use this anywhere else. Can you explain why you think we should do that?

PR #39, opened 2019-01-29
[BPK-2126] Use UTC for calendar RN interface

k0nserv:
Change the calendar to use UTC for the selected dates to align with
Android. Depends on merging https://github.com/Skyscanner/backpack-ios/pull/165 and then releasing `backpack-ios`.

The tests I added will not currently run on CI because we don't have CI set up to run on the macOS infrastructure of Travis. This needs to be changed, but I suggest we do another PR for it

### DIFF
--- a/packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js
+++ b/packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js
@@ -48,7 +48,7 @@ const commonTests = () => {
           <BpkCalendar
             selectionType={SELECTION_TYPES.single}
             {...defaultProps}
-            minDate={new Date(2019, 4, 19)}
+            minDate={new Date(Date.UTC(2019, 4, 19))}
           />,
         )
         .toJSON();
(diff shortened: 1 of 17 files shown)


tiagoengel:
We will need to change stories to provide date props as UTC too. I can do that as part of the android changes if you prefer

k0nserv:
> We will need to change stories to provide date props as UTC too. I can do that as part of the android changes if you prefer

Good point, I'll do that

k0nserv:
The `minDate` and `maxDate` are not correctly handled atm

packages/react-native-bpk-component-calendar/src/BpkCalendar-test.common.js
```diff
@@ -157,9 +160,9 @@ const commonTests = () => {
           <BpkCalendar
             selectionType={SELECTION_TYPES.multiple}
             selectedDates={[
-              new Date(2019, 4, 19),
-              new Date(2019, 4, 21),
-              new Date(2019, 4, 29),
+              new Date(Date.UTC(2019, 4, 19)),
+              new Date(Date.UTC(2019, 4, 21)),
+              new Date(Date.UTC(2019, 4, 29)),
```
georgegillams: My assumption was that this conversion to UTC would happen in the Calendar component's JS so that consumers don't need to worry about it?

k0nserv: No the whole idea is that the public interface in RN should take only UTC dates as input and output. This is to align with Android

PR #40, opened 2019-01-29
[BPK-2212] New social like icon

matthewdavidson:
**Android:**
screenshot_1548755961

iOS:
simulator screen shot - iphone 8 - 2019-01-29 at 09 29 15


ios/Podfile.lock
```diff
@@ -13,7 +13,7 @@ PODS:
   - glog (0.3.5)
   - React (0.57.7):
     - React/Core (= 0.57.7)
-  - react-native-bpk-component-calendar (0.0.1):
+  - react-native-bpk-component-calendar (0.0.2):
```
matthewdavidson: This seems related to previous release of calendar - how do we ensure we keep the demo app Podfile up-to-date?

ios/native.xcodeproj/project.pbxproj
```diff
@@ -344,11 +334,8 @@
 				D66FC4882109D8610063B1A6 /* Main.storyboard in Resources */,
 				13B07FBF1A68108700A75B9A /* Images.xcassets in Resources */,
 				13B07FBD1A68108700A75B9A /* LaunchScreen.xib in Resources */,
-				6A4A4F4B8F7F4269A13BB2A7 /* BpkIcon.ttf in Resources */,
-				C8B831744C5A4256BD1B5010 /* BpkIcon.eot in Resources */,
-				E19140ED6394488C9626FE7A /* BpkIcon.woff in Resources */,
-				8E33D9A4365E4F84826CF0C6 /* iconMapping.json in Resources */,
-				CD97D274772C442994C3117B /* BpkIconIOS.ttf in Resources */,
```
matthewdavidson: I had to manually clean these up :/

package.json
```diff
@@ -85,7 +85,6 @@
     "@storybook/react-native": "^4.1.1",
     "babel-jest": "^23.6.0",
     "babel-plugin-jest-hoist": "^23.2.0",
-    "bpk-svgs": "^6.1.0",
```
matthewdavidson: I removed the top level version as it seems superfluous ... any ideas?

georgegillams: Seems reasonable. If any of our packages are relying on it they should have it themselves

.flowconfig
```diff
@@ -1,5 +1,4 @@
 [ignore]
-.*/packages/.*/node_modules/.*
```
georgegillams: Why is this necessary?

ios/native.xcodeproj/project.pbxproj
georgegillams: Ouch. That seems a little bizarre...

.flowconfig
```diff
@@ -14,7 +13,6 @@ module.name_mapper='\(@storybook/react-native\)' -> '<PROJECT_ROOT>/scripts/stub
 module.name_mapper='\(@skyscanner/react-native-transitiongroup\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
 module.name_mapper='\(react-native-maps\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
 module.name_mapper='\(recyclerlistview\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
-module.name_mapper='\(bpk-react-utils\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
```
georgegillams: Same question here!

.flowconfig
matthewdavidson: Exactly :P thats why i got rid of it.

In all seriousness, this was a remnant from when we were having flow issues back in the day. That line effectively ignored all flow types in dependencies so it's in our interest to remove that. The build still passes so 👍

.flowconfig
matthewdavidson: We no longer depend on `bpk-react-utils` in this project.

PR #43, opened 2019-01-31
Bpk 2220 expose calendar dates utc

georgegillams:
Haven't updated the Readme as that is being done in a separate PR for BPK-2221 already.

### TODO
 - [x] Fix Number interpretation on Android

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,3 +1,5 @@
 # Unreleased
 
-_Nothing yet..._
+ - react-native-bpk-component-calendar:
+   - Changed date prop format to allow numbers representing dates as well.
+   - Added requirement for dates to be specified in UTC.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+# Breaking:
  - react-native-bpk-component-calendar:
-   - Changed date prop format to allow numbers representing dates as well.
+   - Changed date prop format to allow date timestamps in milliseconds, as well as Date object
    - Added requirement for dates to be specified in UTC.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,6 +1,8 @@
 # Unreleased
 
-# Breaking:
- - react-native-bpk-component-calendar:
-   - Changed date prop format to allow date timestamps in milliseconds, as well as Date object
-   - Added requirement for dates to be specified in UTC.
+**Breaking:**
+
+- react-native-bpk-component-calendar:
+  - Changed date prop format to allow date timestamps in milliseconds, as well as Date object
+  - Date props are now expected to be provided in UTC.
+  - `onSelectedDatesChange` callback will now be called with dates in UTC.
--- a/packages/react-native-bpk-component-calendar/src/BpkCalendar.js
+++ b/packages/react-native-bpk-component-calendar/src/BpkCalendar.js
@@ -36,7 +36,7 @@ export type Props = {
 const createOnChangeHandler = memoize(callback => event => {
   if (event.nativeEvent.selectedDates) {
     const convertedDates = event.nativeEvent.selectedDates.map(
-      dt => new Date(dt * 1000),
+      timestamp => new Date(timestamp * 1000),
     );
     if (callback) {
       callback(convertedDates);
--- a/packages/react-native-bpk-component-calendar/src/BpkCalendar.js
+++ b/packages/react-native-bpk-component-calendar/src/BpkCalendar.js
@@ -19,6 +19,7 @@
 /* @flow */
 
 import React from 'react';
+import memoize from 'lodash/memoize';
 
 import {
   commonPropTypes,
@@ -31,6 +32,17 @@ import NativeCalendar from './NativeCalendar';
 export type Props = {
   ...$Exact<CommonProps>,
 };
+
+const createOnChangeHandler = memoize(callback => event => {
+  if (event.nativeEvent.selectedDates) {
+    const convertedDates = event.nativeEvent.selectedDates.map(
+      dt => new Date(dt * 1000),
+    );
+    if (callback) {
+      callback(convertedDates);
+    }
+  }
+});
 
 const BpkCalendar = (props: Props) => {
   const {
@@ -53,7 +65,7 @@ const BpkCalendar = (props: Props) => {
     <NativeCalendar
       minDate={minDate}
       maxDate={maxDate}
-      onChangeSelectedDates={onChangeSelectedDates}
+      onChangeSelectedDates={createOnChangeHandler(onChangeSelectedDates)}
       selectedDates={selectedDates}
       {...rest}
     />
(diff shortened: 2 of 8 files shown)


packages/react-native-bpk-component-calendar/stories.js
```diff
@@ -270,6 +270,18 @@ storiesOf('react-native-bpk-component-calendar', module)
       ]}
     />
   ))
+  .add('With selected dates (UTC number)', () => (
+    <BpkCalendarExample
+      style={styles.calendarOnly}
+      selectionType={SELECTION_TYPES.range}
+      minDate={new Date(Date.UTC(2019, 0, 2))}
```
k0nserv: This seems to be inconsistent with the test description? `new Date(Date.UTC(2019, 0, 2))` is a `Date` not a `number`. Shouldn't this be just `minDate={Date.UTC(2019, 0, 2)}`?

packages/react-native-bpk-component-calendar/src/NativeCalendar.android.js
```diff
@@ -34,27 +33,11 @@ export type Props = {
...
-    }
-  }
-});
-
 const parseDateToNative = date => {
   if (date) {
     // Return as unix timestamp because the only number data type that allows null
     // in the native side is Integer, and the original milliseconds will overflow it.
-    return date.getTime() / 1000;
+    return new Date(date).getTime() / 1000;
```
k0nserv: This tripped me up since `new Date(date)` when `date` is a `Date` is redundant. Can we add a comment to document that?

packages/react-native-bpk-component-calendar/src/NativeCalendar.android.js
```diff
@@ -34,27 +33,11 @@ export type Props = {
...
-        utcDate.getUTCDate(),
-      );
-    });
-    if (callback) {
-      callback(convertedDates);
-    }
-  }
-});
-
 const parseDateToNative = date => {
```
k0nserv: Could we flow type this so it's clear what type `date` is?

packages/react-native-bpk-component-calendar/src/BpkCalendar.js
```diff
@@ -32,6 +33,17 @@ export type Props = {
   ...$Exact<CommonProps>,
 };
 
+const createOnChangeHandler = memoize(callback => event => {
+  if (event.nativeEvent.selectedDates) {
+    const convertedDates = event.nativeEvent.selectedDates.map(
+      dt => new Date(dt * 1000),
```
k0nserv: Would it be clearer if the parameter here is called `timestamp`?

UNRELEASED.md
```diff
@@ -1,3 +1,5 @@
 # Unreleased
 
-_Nothing yet..._
+ - react-native-bpk-component-calendar:
+   - Changed date prop format to allow numbers representing dates as well.
```
tiagoengel: nit: I would change it to "date prop format to allow date timestamps in milliseconds"

packages/react-native-bpk-component-calendar/src/NativeCalendar.android.js
tiagoengel: Isn't the `new Date(date)` also required for IOS given that we now also support date timestamps?

packages/react-native-bpk-component-calendar/stories.js
georgegillams: Good point

packages/react-native-bpk-component-calendar/src/NativeCalendar.android.js
georgegillams: It doesn't seem to be. I think the native handler can deal with a number

k0nserv: `new Date(date)` is required when `typeof date === "number"`. When it's a `Date` this is a NOP I believe, but it's a bit confusing.

image

packages/react-native-bpk-component-calendar/src/NativeCalendar.android.js
```diff
@@ -34,27 +33,13 @@ export type Props = {
...
-      );
-    });
-    if (callback) {
-      callback(convertedDates);
-    }
-  }
-});
-
-const parseDateToNative = date => {
+const parseDateToNative = (date: ?Date) => {
```
tiagoengel: This should be `?(Date | number)`

tiagoengel:
Also, the UTC with selected dates is a bit borked on IOS

kapture 2019-02-01 at 9 39 49

tiagoengel:
About the `Intl` problem it seems to require a polyfill on Android: https://stackoverflow.com/questions/41736735/react-native-and-intl-polyfill-required-on-android-device

Maybe we should not advertise its usage in the readme... or mention the polyfill. 

Another way we can do it is `new Date(utc.getFullYear(), utc.getMonth(), utc.getDate()).toLocaleDateString()`

k0nserv:
That's too bad. How about the following method? Or we could use the one you suggested

```javascript
d.toLocaleDateString('pt-BR', {timeZone: 'UTC'})
```

tiagoengel:
I don't think that works as expected, this is the output I get on chrome when in `America/Sao_Paulo` timezone.

```
new Date(Date.UTC(2019, 01, 01)).toLocaleDateString('pt-BR', { timezone: 'UTC' })
// "31/01/2019"
utc = new Date(Date.UTC(2019, 01, 01))
new Date(utc.getUTCFullYear(), utc.getUTCMonth(), utc.getUTCDate()).toLocaleDateString()
// "01/02/2019"
```

tiagoengel:
That does seem to work too. Sounds good to me.

I do think we should mention how to convert the UTC date to a local date object, not just formatting.

georgegillams:
This is a limitation of the underlying native component. Maybe we should calculate the selected date in the stories so that it's always in view :)

georgegillams:
@tiagoengel where does this happen?
Seems fine when I run it on an Android emulator?
https://github.com/Skyscanner/backpack-react-native/pull/43#issuecomment-459666479

tiagoengel:
That happens just by running `npm run android` with the default emulator. About the date problem is not a limitation of the underlining component AFAIK. `BpkCalendarExample` is spreading the props at the end, meaning is always sending the initially value of `selectedDates` to the calendar, and not the one in the state of that component.

k0nserv:
Couldn't we just assign the state in the constructor?

```javascript
constructor(props) {
  this.state = { selectedDates: props.initiallySelectedDates };
}
```

packages/react-native-bpk-component-calendar/src/NativeCalendar.android.js
```diff
@@ -34,26 +33,15 @@ export type Props = {
...
-  }
-});
-
-const parseDateToNative = date => {
+const parseDateToNative = (date: ?(Date | number)) => {
   if (date) {
     // Return as unix timestamp because the only number data type that allows null
     // in the native side is Integer, and the original milliseconds will overflow it.
+
+    // The consumer-provided date value could be a number, so wrapping it inside a Date constructor in case
```
tiagoengel: This comment can go away now

UNRELEASED.md
```diff
@@ -1,3 +1,6 @@
 # Unreleased
 
-_Nothing yet..._
+# Breaking:
+ - react-native-bpk-component-calendar:
+   - Changed date prop format to allow date timestamps in milliseconds, as well as Date object
+   - Added requirement for dates to be specified in UTC.
```
tiagoengel: We should also add that now the returned dates are also UTC

PR #44, opened 2019-02-04
[BPK-2209] Add docs for the Android bridge

tiagoengel:


packages/react-native-bpk-component-calendar/src/android/build.gradle
```diff
@@ -18,21 +18,23 @@ def safeExtGet(prop, fallback) {
...
+    targetSdkVersion _targetSdkVersion
 
     versionName getVersionFromJsPackage(project)
   }
 }
 
 dependencies {
   compileOnly "com.facebook.react:react-native:${_reactNativeVersion}"
-  implementation "com.github.skyscanner:backpack-android:${_backpackVersion}"
+  implementation "com.github.skyscanner:backpack-android:2.5.0"
```
tiagoengel: Changed to using the minimum version supported. Maven will always get the latest version if we publish it like this, which is what we want. 

The version I published in our internal repo is already using this version (because the root project is using it too) so this doesn't really change anything. Is just more specific and safe than before.

packages/react-native-bpk-component-calendar/README.md
```diff
@@ -11,10 +11,71 @@ npm install react-native-bpk-component-calendar --save-dev
...
+Having installed the NPM package, add the following configuration to gradle:
+
+  1. Define the `react-native-bpk-component-calendar` project in your `settings.gradle` file:
+
+```groovy
+    include ':react-native-bpk-component-calendar'
+    project(':react-native-bpk-component-calendar').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-bpk-component-calendar/src/android')
+```
+
+  2. Add `react-native-bpk-component-calendar` as a dependency in your `build.gradle` file:
```
amolgupta: Can mention which level `build.gradle`. Project or module. In this case it is module.

packages/react-native-bpk-component-calendar/README.md
```diff
@@ -11,10 +11,71 @@ npm install react-native-bpk-component-calendar --save-dev
...
+    minSdkVersion       = 21
+    buildToolsVersion   = "28.0.3"
+}
+```
+
+If you haven't or are using the pre compiled version bellow, it will use the values shown above.
+
+#### Pre compiled version
+
+Alternativelly, if you are Skyscanner employee and have access to the internal artifactory, you can use the pre compiled version directly. Make sure you have the `infrastructure-maven` registry configured and are logged in, then add the following dependency to your `build.gradle` file:
```
amolgupta: Alternatively, pre compiled version is available on skyscanner internal artifactory for developers with access.

PR #51, opened 2019-02-11
[BPK-2236] Fix range render issue in calendar

k0nserv:
Fix an issue were when the calendar initialy rendered with a range
selected the component would render in a broken state.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,4 +1,8 @@
 # Unreleased
 
 _Nothing yet..._
+**Fixed:**
+
+- react-native-calendar:
+  - Fixed a bug were on initial render, on iOS, the calendar would render with broken layout in some cases.
 
--- a/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
+++ b/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
@@ -17,9 +17,14 @@
  */
 
 #import "RCTBPKCalendarManager.h"
+#import <React/RCTBridge.h>
+#import <React/RCTUIManager.h>
+
+
 #import "RCTConvert+RCTBPKCalendar.h"
 #import "RCTBPKCalendar.h"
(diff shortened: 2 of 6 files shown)


packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -45,6 +50,40 @@ - (UIView *)view {
...
+RCT_EXPORT_METHOD(forceRender:(nonnull NSNumber *)reactTag) {
+    [self.bridge.uiManager addUIBlock:
+     ^(__unused RCTUIManager *uiManager, NSDictionary<NSNumber *, UIView *> *viewRegistry){
+         UIView *view = viewRegistry[reactTag];
+
+         if ([view isKindOfClass:[BPKCalendar class]]) {
+             BPKCalendar *calendar = (BPKCalendar *)view;
+             NSArray<NSDate* > *selectedDates = calendar.selectedDates;
+
+             calendar.selectedDates = @[];
```
karanjthakkar: Do we need to empty the calendar once for it to render correctly on the second reload? Is this a temporary fix or is this going to be around?

k0nserv: It seems to be required yeah

k0nserv:
> LGTM 👍 Might be a good idea though to figure out the underlying cause for this hack to be required.

Yeah that's fair, on native iOS if you create the calendar in `viewDidLoad` and then configure it in`viewDidAppear` it doesn't exhibit this behaviour. However if you configure it in `viewDidLoad` the same thing happens. @Szabadember might know more about the potential root cause

UNRELEASED.md
```diff
@@ -1,4 +1,8 @@
 # Unreleased
 
 _Nothing yet..._
```
georgegillams: Can we remove this line (and never add it again)?

k0nserv: I'll clean it up when I release

PR #54, opened 2019-02-15
[BPK-2250] Add 'type' prop to iOS text button

shaundon:
Up to now, all text buttons were being cloned and having `tintColor` and `disabledTintColor` passed in, meaning that internal styling had no effect as those props would override it.

I've tweaked the logic so that `tintColor` and `disabledTintColor` are only passed in if theming has been applied, so that the `type` prop can do internal styling. 

I've also added a new theme prop for colouring primary buttons. I think this is a breaking change, as any consumers using theming will now see a warning that they haven't provided all the theme props.

To do

- [x] Implementation
- [x] Tests
- [x] Update stories
- [ ] Docs (will do as a follow up in the main repo)
- [x] Unreleased
- [x] Screenshot
- [x] Readme
- [x] Theming


packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarTextButtonIOS-test.ios.js
```diff
@@ -81,28 +83,59 @@ describe('iOS', () => {
...
-    it('should resepect "disabledTintColor"', () => {
+    it('should respect "disabledTintColor"', () => {
       const tree = renderer
         .create(
           <BpkNavigationBarTextButtonIOS
             title="Done"
             leading={false}
             onPress={jest.fn()}
             disabledTintColor="red"
+            type={BUTTON_TYPES.primary}
```
shaundon: This is to ensure that `disabledTintColor` takes precedence over any `type` passed in.

packages/react-native-bpk-component-navigation-bar/src/TitleView.js
```diff
@@ -45,11 +45,13 @@ const styles = StyleSheet.create({
...
 const TitleView = (props: {
   title: string | TitleWithIcon,
-  tintColor: string,
+  tintColor: ?string,
   style: any,
 }) => {
   const { title, tintColor, style, ...rest } = props;
 
+  const tintColorFinal = tintColor || colorGray700;
+
```
shaundon: This change means that internal styling will be applied as a fallback if `tintColor` is not passed in, which is the same as how the iOS text button works. This is necessary because in `BpkNavigationBar.ios.js`, `tintColor` now defaults to `null`.

As this file applies to Android too, technically I've put an iOS style in here (`colorGray700`). In practice this isn't an issue because `BpkNavigationBar.android.js` *always* passes in a white `tintColor`, but I can tweak this file to do a platform check and make the default Android colour white, if we prefer to make it safer.

packages/react-native-bpk-component-navigation-bar/README.md
```diff
@@ -159,12 +159,13 @@ import { BpkNavigationBarTextButtonIOS } from 'react-native-bpk-component-naviga
...
-| emphasize | bool     | false    | false         |
-| onPress   | func     | false    | null          |
-| disabled  | bool     | false    | false         |
+| Property  | PropType                     | Required | Default Value |
+| --------- | ---------------------------- | -------- | ------------- |
+| title     | string                       | true     | -             |
+| disabled  | bool                         | false    | false         |
+| emphasize | bool                         | false    | false         |
+| onPress   | func                         | false    | null          |
+| type      | oneOf('default', 'primary')  | false    | primary       |
```
k0nserv: `'default'` should be the default otherwise this becomes a breaking change

packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarTextButtonIOS-test.ios.js
k0nserv: If you feel this needs to be explained should we include a comment in the source?

packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarTextButtonIOS.ios.js
```diff
@@ -50,9 +60,20 @@ const styles = StyleSheet.create({
...
 });
 
+const tintColors = {
+  [BUTTON_TYPES.default]: colorGray700,
+  [BUTTON_TYPES.primary]: colorBlue500,
+};
+
+const disabledTintColors = {
+  [BUTTON_TYPES.default]: colorGray300,
+  [BUTTON_TYPES.primary]: colorBlue300,
```
k0nserv: Have we had design input on this? I feel like this should be gray300 as well

packages/react-native-bpk-component-navigation-bar/README.md
shaundon: That's a typo - it is default. Good spot!

packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarTextButtonIOS.ios.js
shaundon: No, just my assumption. I don't agree but I'll get a designer to decide.

shaundon: Just to confirm that Chris decided to use gray300, so I've done that.

packages/react-native-bpk-component-navigation-bar/README.md
```diff
@@ -113,6 +113,7 @@ PropTypes.shape({
 **iOS:**
 
 + `navigationBarShadowColor`
++ `navigationBarPrimaryColor`
```
k0nserv: This is now a breaking change since a theme that was previously valid is now not valid

UNRELEASED.md
```diff
@@ -1 +1,7 @@
 # Unreleased
+
+**Breaking:**
+
+- react-native-bpk-component-navigation-bar:
+  - New `type` prop for `BpkNavigationBarTextButtonIOS` to change the button colour.
+  - Added `navigationBarPrimaryColor` theme property, applied to `BpkNavigationBarTextButtonIOS` when type is `primary`. This is a breaking change as you must now supply this theme property to theme the component.
```
georgegillams: I think this could be clearer to make consumers know that it is only breaking iff they are applying a theme

shaundon: Thanks - how about now? I added this to the end:

> If you are not using theming, this release is not breaking.

PR #55, opened 2019-02-18
[BPK-2257] Flow weirdness

matthewdavidson:


packages/react-native-bpk-component-button-link/src/common-types.js
```diff
@@ -33,7 +33,7 @@ export const ICON_ALIGNMENTS = {
 
 export type CommonProps = {
   onPress: (event: SyntheticEvent<>) => void,
-  title: string,
+  title: boolean,
```
matthewdavidson: Flow should complain about this.

matthewdavidson:
I pinpointed this due to the use of `withTheme` HOC which has no flow typing.

PR #58, opened 2019-02-21
[No JIRA] Clean up decisions folder

shaundon:


samstarling:
@shaundon You'd better work.

PR #59, opened 2019-02-21
[No JIRA] PR template

shaundon:
I'm piggybacking off of the main repo's code review guidelines to avoid duplicating it, plus the advice in there is all generic.


matthewdavidson:
Why not piggy back off the main one for iOS too?

PR #66, opened 2019-02-27
[BPK-2257] Introduce android specific .flowconfig

matthewdavidson:
Similar to the react native codeabse -> https://github.com/facebook/react-native/blob/master/.flowconfig.android


.flowconfig.android
```diff
@@ -0,0 +1,76 @@
+[ignore]
+; We fork some components by platform
+.*/*[.]ios.js
```
matthewdavidson: This is the key difference between `.flowconfig` and `.flowconfig.android`

shaundon:
Seems sound! Just gonna give it a go locally before I ✅

matthewdavidson:
bitmoji

package.json
```diff
@@ -26,9 +26,11 @@
     "jest:update": "jest --updateSnapshot",
     "jest:watch": "jest --watch",
     "flow": "flow",
```
shaundon:
```suggestion
     "flow": "npm run flow-check-ios && npm run flow-check-android",
```
And then we can keep the `test` line the same as it already was. Doing this means we still have a single command to run *all* Flow things, which is handy.

matthewdavidson: I think we shouldn't do this. `npm run flow` is useful as is because:
- It uses and existing flow server so that subsequent runs are lightning quick (albeit targeting only ios)
- It allows us to do stuff like `npm run flow -- stop` / `npm run flow -- start` (and any other arguments `flow` supports)
- It aligns with the react-native codebase

PR #70, opened 2019-03-06
[BPK-2291] Add theming support to BpkText

shaundon:
* Adds a theme prop `textFontFamily` to BpkText. 
* There are no fonts that are built in to both platforms so for the storybook I'm using Courier on iOS and serif-monospace on Android, which look similar. Consumers are free to use any fonts they want, this is just what we have in our local environment.
* The unrelated snapshots that have changed are where tests were using BpkText directly. All others are unchanged because of shallow rendering.

screen_shot_2019-03-06_at_10 54 09


shaundon:
> Do we know if this has any impact on performance?

I could do some measuring, but is there much we can do? This is the standard way we do theming, if we go down a different route here wouldn't it imply a problem with our overall approach to theming?

We could do something like have a new `BpkThemedText` component but then that means we wouldn't be able to target all existing instances of BpkText.

matthewdavidson:
A quick comparison of 1000 `BpkText` vs 1000 `withTheme(BpkText)` would quell this. The guide on measuring performance on official docs should help: https://facebook.github.io/react-native/docs/performance#profiling

Comparing before and after with regards to memory usage and frame rate should be enough.

shaundon:
@k0nserv @matthewdavidson just did a quick perf test. I made the following two stories:

```javascript
  .add('Perf test', () => (
    <ScrollView>
      {Array(1000)
        .fill()
        .map((k, i) => (
          <BpkText key={i.toString()}>Developers</BpkText>
        ))}
    </ScrollView>
  ))

  .add('Perf test (themed)', () => (
    <BpkThemeProvider theme={themeAttributes}>
      <ScrollView>
        {Array(1000)
          .fill()
          .map((k, i) => (
            <BpkText key={i.toString()}>Developers</BpkText>
          ))}
      </ScrollView>
    </BpkThemeProvider>
  ))
```

Then I loaded each story with the RN profiler open and compared memory usage (iOS), stutters (Android), UI and JS frames per second while rapidly scrolling through the list.

**iOS**

Without theming:
	60fps for both UI and JS
	RAM constant 263-264mb

With theming:
	60fps for both UI and JS
	RAM constant at 265-6mb


**Android**

Without theming:
	~54fps for both UI and JS
	0 stutters
	

With theming:
	~54fps for both UI and JS
	0 stutters

So as far as I can see, performance is identical. I can dive in and use Android and iOS's native profilers if you think it's needed but this felt sufficient.

matthewdavidson:
@shaundon no i think thats more than enough.

PR #71, opened 2019-03-06
[BPK-2269] Add theme attributes render prop component

k0nserv:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,1 +1,15 @@
 # Unreleased
+
+**Added:**
+
+- react-native-bpk-component-text:
+  - Added support for changing the font using a theme prop.
+- react-native-bpk-theming:
+  - Added a new Render Prop component `BpkThemeAttributes` to support arbitrary usage of a theme's `primaryColor` attribute. If rendered outside of a `BpkThemeProvider` with a defined theme `colorBlue500` is used instead.
+    ```javascript
+    const Color = () => (
+      <BpkThemeAttributes>
+        {({ primaryColor }) => <View style={{ backgroundColor: primaryColor }} />}
+      </BpkThemeAttributes>
+    );
+    ```
--- a/packages/react-native-bpk-theming/src/BpkThemeAttributes.js
+++ b/packages/react-native-bpk-theming/src/BpkThemeAttributes.js
@@ -1,0 +1,59 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow */
+
+import { type Node } from 'react';
+import { colorBlue500 } from 'bpk-tokens/tokens/base.react.native';
+
+import { withTheme } from './BpkThemeProvider';
+
+type Attributes = {
+  primaryColor: string,
+};
+
+export type Props = {
+  children: Attributes => Node,
+  theme: { primaryColor: string },
+};
+
+/**
+ * A render prop component that provides certain values
+ * from the nearest `BpkThemeProvider` with suitable defaults
+ * if rendered without a `BpkThemeProvider`.
+ *
+ * Supported attributes:
+ * + `primaryColor`
+ *
+ * @example
+ * <BpkThemeAttributes>
+ *  {({ primaryColor }) => <View style={{ backgroundColor: primaryColor }} />}
+ * </BpkThemeAttributes>
+ *
+ * @param {Object} props `children` prop is used as render prop.
+ * @return {Node} The return value of `props.children`
+ *
+ */
+const BpkThemeAttributes = (props: Props) => {
+  const { theme, children } = props;
+  const primaryColor =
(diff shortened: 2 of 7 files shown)


UNRELEASED.md
```diff
@@ -1 +1,12 @@
 # Unreleased
+
+**Added:**
+
+- react-native-bpk-theming:
+  - Added a new HOC `withPrimaryColor` to support arbitrary usage of a theme's `primaryColor` attribute. If rendered outside of a `BpkThemeProvider` with a defined theme `colorBlue500` is used instead.
+    ```javascript
+    const Color = withPrimaryColor((props) => (
+      <View style={{ backgroundColor: props.primaryColor }} />
+    ));
```
shaundon: Having a code sample is a great move - we should do this more in other PRs really!

packages/react-native-bpk-theming/src/create-primary-color-extractor.js
```diff
@@ -0,0 +1,62 @@
...
+/* @flow */
+import React, { type ComponentType } from 'react';
+
+import { withTheme } from './BpkThemeProvider';
+
+// TODO: Consider creating a single utility for this as it's used in many places
+const getDisplayName = Component =>
+  Component.displayName || Component.name || 'Component';
+
+export default (defaultValue: string) => <
```
tiagoengel: Not sure I understand the advantage of this approach instead of just a function to read the property. I'm not a big fan of HOCs, specially because here we can have `withPrimaryColor` and `withTheme` together, that's a bit unecessary IMO. 

If we have `withTheme` already I think we should just use that and provide a simple function that reads from it. I believe the fact that users will have to wrap the component with `withTheme` is not a big deal as the function will require the theme, and that would make it clear wrapping is required.

Another point is that this approach is not very flexible, adding a `secondaryColor` would be a lot more complex.

k0nserv: I agree on your point about `secondaryColor` the reason we are going for this over just a function where the consumer can pass the `theme` prop from `withTheme` and have the function extract the primary color. The HOC is preferable over this method because:

1. Consumer has to do less work hence there's less risk to do it incorrectly.
2. We maintain abstraction and can change how we do theming without consumers having to care since they are not directly using `withTheme`

k0nserv:
Pushed render prop version in 4c8784f

packages/react-native-bpk-theming/src/BpkThemeAttributes.js
```diff
@@ -0,0 +1,41 @@
...
+import { colorBlue500 } from 'bpk-tokens/tokens/base.react.native';
+
+import { withTheme } from './BpkThemeProvider';
+
+type Attributes = {
+  primaryColor: string,
+};
+
+export type Props = {
+  children: Attributes => void,
```
tiagoengel: Shouldn't this should be `Attributes => Node` ?

matthewdavidson: return type here is `Node` i think

k0nserv: Oh shiet you are right

UNRELEASED.md
```diff
@@ -4,3 +4,12 @@
 
 - react-native-bpk-component-text:
   - Added support for changing the font using a theme prop.
+- react-native-bpk-theming:
+  - Added a new Render Prop component `BpkThemeAttributes` to support arbitrary usage of a theme's `primaryColor` attribute. If rendered outside of a `BpkThemeProvider` with a defined theme `colorBlue500` is used instead.
```
matthewdavidson:
```suggestion
  - Added a new [Render Prop](https://reactjs.org/docs/render-props.html) component `BpkThemeAttributes` to support arbitrary usage of a theme's `primaryColor` attribute. If rendered outside of a `BpkThemeProvider` with a defined theme `colorBlue500` is used instead.
```

packages/react-native-bpk-theming/src/BpkThemeAttributes.js
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @example
+ * <BpkThemeAttributes>
+ *  {({ primaryColor }) => <View style={{ backgroundColor: primaryColor }} />}
+ * </BpkThemeAttributes>
+ *
+ * @param {Object} props `children` prop is used as render prop.
+ * @return {Node} The return value of `props.children`
+ *
+ */
```
matthewdavidson: Whats the deal with JSDoc comment? We don't do this anywhere else...are we now makign a concerted effort to do this? Doesn't flow negate this somewhat?

k0nserv: We've been doing them in Objective-C and Java/Kotlin which also has a type system. One doesn't preclude the other I think. Most editors will show this extra information when autocompleting which is nice imo

matthewdavidson: Cool!

I'm a little concerned about them rotting but they are no different to our current `README.md`'s in that regard. I guess its yet another place we have to update too...

Change is hard 😅

matthewdavidson: This might be helpful https://www.npmjs.com/package/eslint-plugin-jsdoc

packages/react-native-bpk-theming/src/BpkThemeAttributes.js
```diff
@@ -0,0 +1,58 @@
...
+type Attributes = {
+  primaryColor: string,
+};
+
+export type Props = {
+  children: Attributes => Node,
+  theme: { primaryColor: string },
+};
+
+/** A render prop component that provides certain values
```
matthewdavidson: Also this is triggering me 😅, surely it should be:
```
/** 
 * A render prop component that provides certain values
 * from the nearest `BpkThemeProvider` with suitable defaults
 * if rendered without a `BpkThemeProvider`.

```

packages/react-native-bpk-theming/src/BpkThemeAttributes.js
k0nserv: We already have something in our eslint config that enforces writing JSdoc comments correctly however it only triggers when you start writing one

PR #76, opened 2019-03-12
[BPK-2299] Theming support for flat list and section list

shaundon:
* The section list and flat list components used to duplicate a lot of files (the list item, separator and the radio icon along with associated tests etc). This makes flatlist a dependency of sectionlist so that we can reuse the same code. All the exports remain the same so it's not breaking and won't change anything for consumers.
* Rather than make `BpkRadioIcon.android` (internal component) themeable I made it accept a `tintColor` prop so that we don't have to repeat the theming logic we're already doing in the list item component.

Screenshot_1552398770
Screenshot_1552398480
Simulator Screen Shot - iPhone 8 - 2019-03-12 at 13 37 14
Simulator Screen Shot - iPhone 8 - 2019-03-12 at 13 32 01


packages/react-native-bpk-component-flat-list/src/BpkRadioIcon.android.js
```diff
@@ -75,6 +67,7 @@ const BpkRadioIcon = ({ selected }: Props) => {
 
 BpkRadioIcon.propTypes = {
   selected: PropTypes.bool,
+  tintColor: PropTypes.string.isRequired,
```
k0nserv: This is a breaking change strictly. Could we default it to `colorBlue500` to prevent that?

packages/react-native-bpk-component-flat-list/src/theming.js
```diff
@@ -16,10 +16,12 @@
  * limitations under the License.
  */
 
+// If theming is ever expanded to support other types, this should be changed
+// to something akin to BpkButton's theming functions.
 /* @flow */
+import { makeThemePropType } from 'react-native-bpk-theming';
 
-import commonTests from './BpkSectionListItem-test.common';
+const REQUIRED_THEME_ATTRIBUTES: Array<string> = ['flatListSelectedItemColor'];
```
k0nserv: nitpic: The `Array<string>` type is redundant here as flow will infer it quite easily

packages/react-native-bpk-component-flat-list/src/BpkRadioIcon.android.js
shaundon: This component is only used inside the component and isn't exported to consumers, so I don't think this is breaking?

packages/react-native-bpk-component-flat-list/src/theming.js
shaundon: It's copypasta from `BpkText` - do you think it's worth changing?

packages/react-native-bpk-component-flat-list/src/BpkRadioIcon.android.js
shaundon: I just realised how ridiculous this sounds 😂 

> This component is only used inside the component

I mean it's only used by flatlist lol

k0nserv: Right so we never export it at all?

packages/react-native-bpk-component-flat-list/src/theming.js
k0nserv: Yeah :+1

shaundon: Cool, I'll change it everywhere then for consistency 🙌

packages/react-native-bpk-component-flat-list/src/BpkRadioIcon.android.js
shaundon: That's right - here's `index.js` for this package: https://github.com/Skyscanner/backpack-react-native/blob/master/packages/react-native-bpk-component-flat-list/index.js

packages/react-native-bpk-component-flat-list/src/theming.js
shaundon: All done, here's the specific commit: https://github.com/Skyscanner/backpack-react-native/pull/76/commits/eb10d34f9d44348c9e333fe7488330f34aede110

packages/react-native-bpk-component-flat-list/src/BpkFlatListItem.android.js
```diff
@@ -70,7 +73,23 @@ class BpkFlatListItem extends React.PureComponent<FlatListItemProps> {
...
+      textStyles.push(styles.textSelected);
+      const themeAttributes = getThemeAttributes(
+        REQUIRED_THEME_ATTRIBUTES,
+        theme,
+      );
+      if (themeAttributes) {
+        textStyles.push({
+          color: themeAttributes.flatListSelectedItemColor,
+        });
+        tintColorFinal = themeAttributes.flatListSelectedItemColor;
```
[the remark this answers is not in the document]
georgegillams: Just to make it a bit more readable ;)

PR #77, opened 2019-03-12
[BPK-2295] Upgrade React Native to 0.59.0

shaundon:
## Tasks
- [x] Upgrade RN
- [x] Run on iOS
- [x] Run on Android
- [x] Confirm tests are ok
- [x] Upgrade Flow to align with RN 0.59.0's version
- [x] Go through each story and note anything broken (iOS)
- [x] Go through each story and note anything broken (Android)
- [x] Upgrade the React and RN versions in the `package.json` of each component

## Issues identified so far
- [x] Jest doesn't run (fixed by adding `testMatch: null`)
- [x] Linear gradient doesn't work (fixed by upgrading to latest)
- [x] Flow issues (legit errors, fixed manually)
- [x] Calendar is redboxing on iOS (fixed by enforcing numbers instead of dates sent over the bridge)
- [ ] Async storage warning (this is [an issue with Storybook](https://github.com/storybooks/storybook/issues/6078), I've made [a PR](https://github.com/storybooks/storybook/pull/6079)). I think this is acceptable as-is - it just displays a yellow box error when Storybook first loads and I anticipate the fix being merged and released soon.
- [x] Doesn't run on Android (`> Could not get unknown property 'mergeResourcesProvider' for object of type com.android.build.gradle.internal.api.ApplicationVariantImpl.`) (Fixed by doing this https://github.com/wix/react-native-navigation/issues/4757#issuecomment-468133753)


package.json
```diff
@@ -102,10 +102,10 @@
...
     "prompt": "^1.0.0",
     "prop-types": "^15.5.10",
-    "react": "16.6.1",
-    "react-dom": "16.6.1",
-    "react-native": "0.57.7",
-    "react-native-linear-gradient": "react-native-community/react-native-linear-gradient#20928ce2ac8038e8545c7f6c3faab05f866deabd",
+    "react": "16.6.3",
+    "react-dom": "16.6.3",
+    "react-native": "0.58.5",
+    "react-native-linear-gradient": "^2.5.3",
```
shaundon: `ColorPropType` was [deprecated in 0.58](https://github.com/react-native-community/react-native-linear-gradient/issues/356) so I upgraded `react-native-linear-gradient` to the latest version, which seems to fix the problem, and gets us back on a normal version and not this branch 🙌

packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -56,6 +59,7 @@ const styles = StyleSheet.create({
 
 export type Props = {
   ...$Exact<ImageProps>,
+  style: ViewStyleProp,
```
shaundon: Flow was giving us an error here because `Image.style` isn't _quite_ the same as `View.style`, but we were using `Image`'s `style` type here then applying it to a view (on line 144). So, I've made it explicitly use the `style` type from `View`.

package.json
```diff
@@ -92,7 +92,7 @@
     "eslint-config-skyscanner-with-prettier": "^0.8.0",
     "eslint-plugin-flowtype": "^3.2.0",
     "eslint_d": "^7.2.0",
-    "flow-bin": "^0.78.0",
+    "flow-bin": "^0.86.0",
```
shaundon: Upgraded Flow to 86 to match the version used in RN 0.58.5: https://github.com/facebook/react-native/blob/v0.58.5/.flowconfig#L101

.flowconfig
```diff
@@ -24,7 +24,6 @@
 [libs]
 node_modules/react-native/Libraries/react-native/react-native-interface.js
 node_modules/react-native/flow/
-node_modules/react-native/flow-github/
```
shaundon: Removed to fix this notice:

Screen Shot 2019-03-13 at 17 57 04

package.json
```diff
@@ -49,6 +49,7 @@
     "preset": "react-native",
     "verbose": true,
     "testRegex": ".*-test(\\.ios)?(\\.android)?\\.js$",
+    "testMatch": null,
```
shaundon: Fixes a Jest issue:

https://github.com/kulshekhar/ts-jest/issues/756#issuecomment-424057361

package.json
```diff
@@ -83,16 +84,16 @@
   "devDependencies": {
     "@storybook/addon-actions": "^4.1.1",
     "@storybook/addon-links": "^4.1.1",
-    "@storybook/react-native": "^4.1.1",
-    "babel-jest": "^23.6.0",
-    "babel-plugin-jest-hoist": "^23.2.0",
+    "@storybook/react-native": "^4.1.14",
+    "babel-jest": "^24.5.0",
+    "babel-plugin-jest-hoist": "^24.3.0",
```
shaundon: I updated these to try and fix another [issue](https://github.com/Skyscanner/backpack-react-native/pull/77/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.kt&file-filters%5B%5D=.snap#r265502545) but the upgrade doesn't seem to break anything so may as well do it.

package.json
```diff
@@ -83,16 +84,16 @@
...
     "bpk-tokens": "^27.4.7",
     "colors": "^1.3.3",
     "danger": "^7.0.11",
     "eslint-config-skyscanner-with-prettier": "^0.8.0",
     "eslint-plugin-flowtype": "^3.2.0",
     "eslint_d": "^7.2.0",
-    "flow-bin": "^0.78.0",
+    "flow-bin": "^0.92.0",
```
shaundon: To align with RN 0.59's Flow version.

android/build.gradle
```diff
@@ -9,7 +9,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.0'
+        classpath 'com.android.tools.build:gradle:3.3.2'
```
shaundon: To fix a build issue (see https://github.com/wix/react-native-navigation/issues/4757#issuecomment-468133753)

package.json
```diff
@@ -102,20 +103,19 @@
...
   },
   "greenkeeper": {
     "ignore": [
       "flow-bin",
       "react",
       "react-dom",
       "react-test-renderer",
-      "react-native",
-      "react-native-linear-gradient"
+      "react-native"
```
shaundon: We no longer need to ignore linear gradient.

packages/react-native-bpk-component-calendar/src/NativeCalendar.ios.js
```diff
@@ -59,7 +70,7 @@ class BpkCalendar extends Component<Props, {}> {
       const handle = findNodeHandle(this.calendarRef.current);
       UIManager.dispatchViewManagerCommand(
         handle,
-        UIManager.RCTBPKCalendar.Commands.forceRender,
+        UIManager.getViewManagerConfig('RCTBPKCalendar').Commands.forceRender,
```
shaundon: To remove a deprecation warning. See https://github.com/facebook/react-native/blob/d6236796b285e6ad19c53c5308a0ad9c10792a05/Libraries/ReactNative/UIManagerProperties.js#L15

packages/react-native-bpk-component-calendar/src/NativeCalendar.ios.js
```diff
@@ -86,10 +97,10 @@ class BpkCalendar extends Component<Props, {}> {
     return (
       <RCTBPKCalendar
         ref={this.calendarRef}
-        minDate={minDate}
-        maxDate={maxDate}
+        minDate={parseDateToNative(minDate)}
+        maxDate={parseDateToNative(maxDate)}
         onDateSelection={onChangeSelectedDates}
-        selectedDates={selectedDates}
+        selectedDates={selectedDates.map(parseDateToNative)}
```
shaundon: This was a nightmare issue to fix. Essentially something appears to have changed on the RN iOS side that means dates sent over the bridge weren't making it over properly, so it was failing to parse and causing red box errors.

I never figured out what exactly changed internally to make this happen, but I fixed it by taking inspiration from the Android version of this component and cast everything to a timestamp before sending it over the bridge, which makes it work.

packages/react-native-bpk-component-calendar/src/android/src/main/java/net/skyscanner/backpack/reactnative/calendar/CalendarPackage.kt
```diff
@@ -21,14 +21,15 @@ import com.facebook.react.ReactPackage
...
 
 class CalendarPackage : ReactPackage {
 
-  override fun createNativeModules(reactContext: ReactApplicationContext?): MutableList<NativeModule> {
+  override fun createNativeModules(reactContext: ReactApplicationContext): MutableList<NativeModule> {
     return mutableListOf()
   }
 
-  override fun createViewManagers(reactContext: ReactApplicationContext?): MutableList<ViewManager<*, *>> {
+  override fun createViewManagers(reactContext: ReactApplicationContext): MutableList<ViewManager<*, *>> {
```
shaundon: After upgrading Gradly things I started seeing build errors here that we were failing to override the methods we were supposed to. I checked and it appears the signatures changed so that `ReactApplicationContext` is no longer optional. This change doesn't appear to break anything.

packages/react-native-bpk-theming/stories.js
```diff
@@ -109,10 +109,12 @@ class BpkThemePicker extends Component<{}, State> {
   }
 
   switchTheme = value => {
-    this.setState({
-      themeId: value,
-      theme: this.themes[value],
-    });
+    if (typeof value === 'string') {
```
shaundon: Ridiculous check to stop Flow mithering me

packages/react-native-bpk-component-button-link/src/common-types.js
```diff
@@ -21,6 +21,7 @@
 import { type Node } from 'react';
 import PropTypes from 'prop-types';
 import { makeThemePropType } from 'react-native-bpk-theming';
+import type { PressEvent } from 'react-native/Libraries/Types/CoreEventTypes';
```
georgegillams: Are you meant to use stuff like this? Isn't there a danger that they change what's exposed as it's not part of the official API?

shaundon: There is that risk, but I don't see another way of doing it as it doesn't appear to be exported on the public interface, and if it breaks we'll find out as soon as we try to upgrade RN and can change it.

I did try this:

```
type TouchableProps = ElementProps<typeof TouchableWithoutFeedback>;
type TouchableOnPressProp = $PropertyType<TouchableProps, 'onPress'>;
```

But I get a Flow error and I couldn't figure it out:

```
Cannot instantiate $PropertyType because property onPress is missing in mixed [1].

     29│ } from './customPropTypes';
     30│
     31│ type TouchableProps = ElementProps<typeof TouchableWithoutFeedback>;
 [1] 32│ type TouchableOnPressProp = $PropertyType<TouchableProps, 'onPress'>;
```

@matthewdavidson may have an opinion as he's done similar things in his Flow-ing PRs.

matthewdavidson: You can use `SyntheticEvent<>`, no?

shaundon: It doesn't like that anymore, that's what we were using before this PR:

Screen Shot 2019-03-14 at 14 33 34

.flowconfig
matthewdavidson: Can we also have a look at the `.flowconfig` of RN at 0.59.0? That might contain some new configuration that we should also add.

shaundon: I was going to do that, but if the config we have is already working for us, what is there to be gained by changing it?

matthewdavidson: We used that mentality previously, only to find out that we were missing out on a bunch of flow functionality. 

In other words, when we aligned our version of flow to RN's version and subsequently aligned the `.flowconfig`, we found errors we were otherwise oblivious to. 

Who knows, perhaps https://github.com/Skyscanner/backpack-react-native/pull/77#discussion_r265600509 goes away when we align the config?

shaundon: That's a fair point! Let me investigate 🤓

shaundon: I've updated it based on RN's latest configs, no changes it seems (including https://github.com/Skyscanner/backpack-react-native/pull/77#discussion_r265600509 unfortunately) but as you say, good to be aligned with the latest and greatest.

package.json
matthewdavidson: Nitpick: maybe this should explicitly be `"0.92.0"`.

Fully appreciate that I probably set the carat last time haha

shaundon: Haha - well I went with the carat because that's [what RN does](https://github.com/facebook/react-native/blob/v0.59.0/package.json#L230) in their repo but I'm happy to lock it if you feel strongly about it.

matthewdavidson: No lets follow them :)

PR #78, opened 2019-03-15
[No JIRA] Check spelling in markdown files

shaundon:
When I contributed to Storybook I noticed they had a `.spelling` file so I investigated. I think we can make use of the same library, so I've added it and integrated it into `npm test`. It's fixed quite a lot of spelling errors so it's providing value already.

One downside is that I've had to add **a lot** of exclusions, for things like component names and tech terms. I'd like to make a PR to the main repo to allow it to optionally ignore camelCase and PascalCase words so we could remove much of these, but it looks as though it isn't maintained very much, which is definitely a downside.

If this is successful we can add it to our other repos.

Thoughts?


shaundon:
> Is the danger warning coming from this new lib?

I have no idea! Let me look into that 🤔

PR #79, opened 2019-03-15
[NO-JIRA] Fix package-lock

tiagoengel:
This only changes package-lock*s* that have changed during the install phase.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "android": "npm run android:emulator & npm run android:build",
     "spellcheck": "mdspell -r --en-gb --ignore-acronyms --ignore-numbers --no-suggestions '*.md' '**/*.md' '!**/node_modules/**/*.md' '!ios/Pods/**/*.md'",
     "spellcheck:interactive": "mdspell --en-gb --ignore-acronyms --ignore-numbers '*.md' '**/*.md' '!**/node_modules/**/*.md' '!ios/Pods/**/*.md'",
-    "start": "react-native start"
+    "start": "react-native start",
+    "fix:package-lock": "git status -s | grep package-lock.json | awk '{print $2}' | xargs replace --silent 'http://' 'https://'"
   },
   "jest": {
     "preset": "react-native",
@@ -109,7 +110,8 @@
     "react-dom": "16.6.1",
     "react-native": "0.57.7",
     "react-native-linear-gradient": "react-native-community/react-native-linear-gradient#20928ce2ac8038e8545c7f6c3faab05f866deabd",
-    "react-test-renderer": "16.6.1"
+    "react-test-renderer": "16.6.1",
+    "replace": "^1.1.0"
   },
   "greenkeeper": {
     "ignore": [
--- a/package-lock.json
+++ b/package-lock.json
@@ -3320,7 +3320,7 @@
     },
     "@storybook/podda": {
       "version": "1.2.3",
-      "resolved": "http://registry.npmjs.org/@storybook/podda/-/podda-1.2.3.tgz",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -45,7 +45,8 @@
     "android": "npm run android:emulator & npm run android:build",
     "spellcheck": "mdspell -r --en-gb --ignore-acronyms --ignore-numbers --no-suggestions '*.md' '**/*.md' '!**/node_modules/**/*.md' '!ios/Pods/**/*.md'",
     "spellcheck:interactive": "mdspell --en-gb --ignore-acronyms --ignore-numbers '*.md' '**/*.md' '!**/node_modules/**/*.md' '!ios/Pods/**/*.md'",
-    "start": "react-native start"
+    "start": "react-native start",
+    "fix:package-lock": "git status -s | grep package-lock.json | awk '{print $2}' | xargs sed -i 's/http:\\/\\//https:\\/\\//g'"
```
georgegillams: The `-i` expects input

```suggestion
    "fix:package-lock": "git status -s | grep package-lock.json | awk '{print $2}' | xargs sed -i '' 's/http:\\/\\//https:\\/\\//g'"
```

georgegillams: At the moment it is interpretting the regex as the input for `-i` and doing this

Screenshot 2019-03-15 at 15 45 34

tiagoengel: Are you sure? `-i` as far as I know is `in-place`, meaning to replace the file

tiagoengel: Ah, nvm this is that problem with mac os using an old version of sed

georgegillams:
```
    -i extension
             Edit files in-place, saving backups with the specified extension.  If a zero-length extension is given, no backup will be saved.  It is not recommended to give a zero-length extension when in-place editing files, as you risk corruption or partial content in situations where disk space is exhausted, etc.
```

tiagoengel: Yeah, but that is not the behaviour with GNU seed (which I have locally and is by default on linux) and they are not compatible

tiagoengel: So I've replaced this with `replace` (no pun intended) and it works now. I had to remove from post install though as it is not working. I suspect the package lock is only saved on the disk after the post install script runs

package.json
```diff
@@ -45,7 +45,8 @@
     "android": "npm run android:emulator & npm run android:build",
     "spellcheck": "mdspell -r --en-gb --ignore-acronyms --ignore-numbers --no-suggestions '*.md' '**/*.md' '!**/node_modules/**/*.md' '!ios/Pods/**/*.md'",
     "spellcheck:interactive": "mdspell --en-gb --ignore-acronyms --ignore-numbers '*.md' '**/*.md' '!**/node_modules/**/*.md' '!ios/Pods/**/*.md'",
-    "start": "react-native start"
+    "start": "react-native start",
+    "fix:package-lock": "git status -s | grep package-lock.json | awk '{print $2}' | xargs replace --silent 'http://' 'https://'"
```
georgegillams: We should make this run this in `postinstall` - but that won't work on CI as Travis doesn't actually run within the git repo.
So we would need something like this to make Travis happy with running as a postinstall. This command is becomming an absolute unit 😆

```suggestion
    "fix:package-lock": "(git rev-parse --is-inside-work-tree && git status -s | grep package-lock.json | awk '{print $2}' | xargs replace --silent 'http://' 'https://') || true",
```

georgegillams: Alternatively we make it a pre-commit hook, or we just skip the git check, so then we don't need to worry about CI 🤷‍♂️

shaundon:
This is really really old, can we close it?

olliecurtis:
I think we can close this as its been open for a very long time and no longer seems to be an issue

PR #82, opened 2019-03-18
[BPK-2298] Add textInputFocusedColor theme prop

tiagoengel:

### DIFF
--- a/packages/react-native-bpk-component-text-input/README.md
+++ b/packages/react-native-bpk-component-text-input/README.md
@@ -120,4 +120,5 @@ More details about how to use and define a mask can be found here: https://githu
 
 ## Theme Props
 
+* `textFontFamily`
 * `textInputFocusedColor`
--- a/packages/react-native-bpk-component-text-input/README.md
+++ b/packages/react-native-bpk-component-text-input/README.md
@@ -117,3 +117,7 @@ export default class App extends Component {
 ## Mask
 
 More details about how to use and define a mask can be found here: https://github.com/benhurott/tinymask
+
+## Theme Props
+
+* `textInputFocusedColor`
--- a/packages/react-native-bpk-component-text-input/package.json
+++ b/packages/react-native-bpk-component-text-input/package.json
@@ -16,8 +16,8 @@
   "dependencies": {
     "bpk-tokens": "^27.4.7",
     "prop-types": "^15.5.8",
-    "react-native-bpk-component-icon": "^1.14.4",
-    "react-native-bpk-component-text": "^4.1.2",
+    "react-native-bpk-component-icon": "^1.14.5",
+    "react-native-bpk-component-text": "^4.1.3",
     "react-native-bpk-theming": "^1.1.2",
     "tinymask": "^1.0.2"
   }
--- a/packages/react-native-bpk-component-text-input/package.json
+++ b/packages/react-native-bpk-component-text-input/package.json
@@ -16,8 +16,9 @@
   "dependencies": {
     "bpk-tokens": "^27.4.7",
     "prop-types": "^15.5.8",
-    "react-native-bpk-component-icon": "^1.14.5",
-    "react-native-bpk-component-text": "^4.1.3",
+    "react-native-bpk-component-icon": "^1.14.4",
+    "react-native-bpk-component-text": "^4.1.2",
+    "react-native-bpk-theming": "^1.1.2",
     "tinymask": "^1.0.2"
   }
 }
--- a/packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
+++ b/packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
@@ -21,6 +21,7 @@
 import React from 'react';
 import TestRenderer from 'react-test-renderer';
 import { StyleSheet, View } from 'react-native';
+import BpkThemeProvider from 'react-native-bpk-theming';
 import { spacingSm } from 'bpk-tokens/tokens/base.react.native';
 
 import BpkTextInput from './BpkTextInput';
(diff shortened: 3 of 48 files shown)


packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
```diff
@@ -178,6 +179,18 @@ const commonTests = () => {
 
       expect(testRenderer.toJSON()).toMatchSnapshot();
     });
+
+    it('should support theming', () => {
```
tiagoengel: This is not the a very good test as the style here is an animated value and only changes if the input is focused (and after the animation). We don't currently test any of this (what happens when it receives focus). I decided to follow this as adding a proper test will require a lot of work (and probably adding enzyme)

packages/react-native-bpk-component-text-input/package.json
```diff
@@ -16,8 +16,9 @@
   "dependencies": {
     "bpk-tokens": "^27.4.7",
     "prop-types": "^15.5.8",
-    "react-native-bpk-component-icon": "^1.14.5",
-    "react-native-bpk-component-text": "^4.1.3",
+    "react-native-bpk-component-icon": "^1.14.4",
+    "react-native-bpk-component-text": "^4.1.2",
```
shaundon: It looks like these versions have regressed - icon should be `^1.1.14` and text should be `^4.1.3`.

packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
shaundon: Yeah that's fair enough. Perhaps we should have a small follow up card to do that. What do you think?

shaundon:
Android screenshot:

Screenshot_1552983545

shaundon:
iOS screenshot:

Simulator Screen Shot - iPhone 8 - 2019-03-19 at 08 27 33

shaundon:
The text in the themed example should probably change to use the `textFontFamily` theme prop too. In other examples this happens automatically because `BpkText` is in there, but here we use `Animated.Text` and `TextInput` so it doesn't.

(Note: the validation message **does** get themed because that uses `BpkText`).

As part of this PR I think we should add `textFontFamily` to the theme props and apply them to the `Animated.Text` and `TextInput` styles. You should be able to import `textFontFamily` from `BpkText` so there's no duplication too.

packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
tiagoengel: Not sure how beneficial this would be to be honest, the cost/benefit of maintaining such a test doesn't seem good enough. Also we don't have many issues with this component as far as I know

tiagoengel:
Fair enough, I'll add the textFontFamily prop too

packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
shaundon: Alright cool 👍

tiagoengel:
Looks like this now:
Screenshot_1552989162

packages/react-native-bpk-component-text-input/README.md
```diff
@@ -116,4 +116,9 @@ export default class App extends Component {
 
 ## Mask
 
-More details about how to use and define a mask can be found here: https://github.com/benhurott/tinymask
\ No newline at end of file
+More details about how to use and define a mask can be found here: https://github.com/benhurott/tinymask
+
+## Theme Props
+
+* `textFontFamily`
```
shaundon:
```suggestion
* `textFontFamily` (inherited from [`BpkText`](/components/text/?platform=native)
```

Nitpick but I think it'll make it clearer to users.

packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -185,18 +195,36 @@ class BpkTextInput extends Component<Props, State> {
...
+      focusedColor,
     );
 
+    const inputTextStyle = [styles.input];
+
+    if (themeAttributes) {
+      inputTextStyle.push({ fontFamily: themeAttributes.textFontFamily });
+      animatedLabelStyle.push({ fontFamily: themeAttributes.textFontFamily });
+    }
+
```
shaundon: We can avoid the double check of `themeAttributes` if we move this `if` block upwards and then add the redefinition of `focusedColor` to it, like:

```
let focusedColor = undefined;
if (themeAttributes) {
      inputTextStyle.push({ fontFamily: themeAttributes.textFontFamily });
      animatedLabelStyle.push({ fontFamily: themeAttributes.textFontFamily });
      focusedColor = themeAttributes.textInputFocusedColor;
}
```

The optimisation is minor but it'll help make it clearer if all the theme logic is in the same place.

packages/react-native-bpk-component-text-input/src/theming.js
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @flow
+
+import { makeThemePropType } from 'react-native-bpk-theming';
+
+const REQUIRED_THEME_ATTRIBUTES = ['textFontFamily', 'textInputFocusedColor'];
```
shaundon: Perhaps we should export `REQUIRED_THEME_ATTRIBUTES` from `BpkText` and then import it here so it wouldn't need to be duplicated, or do you think that's overkill?

packages/react-native-bpk-component-text-input/src/BpkTextInput.js
tiagoengel: We can't because the function that returns `animatedLabelStyle` depends on `focusedColor`

shaundon: Ahh - never mind then 😄

packages/react-native-bpk-component-text-input/src/theming.js
tiagoengel: I don't think we should at this point, we might want to have different props for text and input. Also even if we import it here we would still need to make use of them in the code so it wouldn't give us any benefit right now.

shaundon: Ok fair enough 👍

PR #83, opened 2019-03-18
Integrate RCTBPKCalendar with SimpleDate

zsoltvarnai:
The update integrates RCTBPKCalendar with the new interface of BPKCalendar, which is using SimpleDates.

### DIFF
--- a/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
+++ b/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
@@ -68,28 +68,31 @@ NS_ASSUME_NONNULL_BEGIN
     NSDate *localDate = [RCTBPKCalendarDateUtils convertDateToLocal:minDate
                                                       localCalendar:self.gregorian
                                                         utcCalendar:self.utcCalendar];
-    [super setMinDate:localDate];
+    BPKSimpleDate *simpleDate = [super simpleDateFromDate:localDate];
+    [super setMinDate:simpleDate];
 }
 
 - (void)setMaxDate:(NSDate *)maxDate {
     NSDate *localDate = [RCTBPKCalendarDateUtils convertDateToLocal:maxDate
                                                       localCalendar:self.gregorian
                                                         utcCalendar:self.utcCalendar];
-    [super setMaxDate:localDate];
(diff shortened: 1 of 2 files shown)


packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
```diff
@@ -68,28 +68,31 @@ - (void)setMinDate:(NSDate *)minDate {
     NSDate *localDate = [RCTBPKCalendarDateUtils convertDateToLocal:minDate
                                                       localCalendar:self.gregorian
                                                         utcCalendar:self.utcCalendar];
-    [super setMinDate:localDate];
+    BPKSimpleDate *simpleDate = [super simpleDateFromDate:localDate];
```
k0nserv: Should we consider having both a `simpleDateFromLocalDate` and `simpleDataFromUTCDate`?

k0nserv: If so we should include this in https://github.com/Skyscanner/backpack-ios/pull/203

zsoltvarnai: Possibly, but the "LocalDate" naming might cause confusion if used on the public interface. In the Skyscanner app we are calling the timezone independent dates LocalDate and its not just a custom name, we are following the naming of java.Time.LocalDate:
https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html
"A date without a time-zone in the ISO-8601 calendar system"

So technically this would be something like simpleDateFromDateInLocalTimeZone, not sure there is a more compact way to name it. :)
What do you think, would you still change the current simpleDateFromDate to these 2?

k0nserv: Nah let's keep it as is and use the conversion from UTC to local time zone here before calling `simpleDateFromDate:`

georgegillams:
Just one minor thing - could you add an entry to `UNRELEASED.md`, following the style seen in [CHANGELOG.md](https://github.com/Skyscanner/backpack-ios/blob/master/CHANGELOG.md)?

k0nserv:
Closing this to reopen as non-fork.

PR #91, opened 2019-03-25
[BPK-2345]  Align `valid` selects to text input behaviour

matthewdavidson:
We are removing the green bottom border from selects in order to align with text inputs, see https://backpack.github.io/components/text-input?platform=native.

| Before      | After |
| ----------- | ----------- |
| android-before | android-after |
| ios-before | ios-after | 

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,4 +1,4 @@
 # Unreleased
 
 - react-native-bpk-component-select
-  - `valid` selects now align to behaviour on text inputs with regards to border colour.
+  - `BpkSelect` with `valid={true}` no longer has a green border to match the behaviour of `BpkTextInput` which also doesn't have one. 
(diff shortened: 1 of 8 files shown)


UNRELEASED.md
```diff
@@ -1 +1,4 @@
-# Unreleased
\ No newline at end of file
+# Unreleased
+
+- react-native-bpk-component-select
+  - `valid` selects now align to behaviour on text inputs with regards to border colour.
```
shaundon: I think this can be clearer to users about what we've changed:

```suggestion
  - `BpkSelect` with `valid={true}` no longer has a green border to match the behaviour of `BpkTextInput` which also doesn't have one. 
```

matthewdavidson:
@shaundon i updated the unreleased, can you re-approve? 😇

PR #96, opened 2019-04-02
[No JIRA] Increase tap area for iOS nav bar buttons

k0nserv:

### DIFF
--- a/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarBackButtonIOS.ios.js
+++ b/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarBackButtonIOS.ios.js
@@ -35,14 +35,19 @@ export type Props = {
   tintColor: ?string,
 };
 
+// NOTE: this file explicitly does not use the Backpack tokens(for spacing) because it's based on UIKit design tokens not Backpack.
 const styles = StyleSheet.create({
   button: {
     flexDirection: 'row',
     zIndex: 2,
     alignItems: 'center',
+    height: '100%',
   },
   backIcon: {
     marginEnd: 8, // eslint-disable-line backpack/use-tokens
+  },
+  backIconWithoutTitle: {
+    marginEnd: 16, // eslint-disable-line backpack/use-tokens
   },
 });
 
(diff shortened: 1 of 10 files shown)


packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarBackButtonIOS.ios.js
```diff
@@ -40,6 +40,7 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     zIndex: 2,
     alignItems: 'center',
+    height: '100%',
```
georgegillams: Doesn't this need `justifyContent: 'center'` too?

packages/react-native-bpk-component-icon/package-lock.json
```diff
@@ -17,12 +17,6 @@
 				"color": "^3.0.0"
 			}
 		},
-		"brcast": {
```
georgegillams: Why did `package-lock.json` change?

packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarBackButtonIOS.ios.js
k0nserv: Nope, it's a row and already has `aligntItems: 'center'`

packages/react-native-bpk-component-icon/package-lock.json
k0nserv: No idea

packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBarBackButtonIOS.ios.js
```diff
@@ -40,16 +40,21 @@ const styles = StyleSheet.create({
...
   },
   backIcon: {
     marginEnd: 8, // eslint-disable-line backpack/use-tokens
   },
+  backIconWithoutTitle: {
+    marginEnd: 16, // eslint-disable-line backpack/use-tokens
```
georgegillams: What's the reason for not using tokens? I think there should be a comment to go with this explaining the decision...

k0nserv: Sorted

packages/react-native-bpk-component-icon/package-lock.json
georgegillams: Do you wanna revert it then? 😆

shaundon: Yeah we should probably revert this

PR #101, opened 2019-04-12
Bpk 2384 icon alignment

ojcurtis:


shaundon:
This isn't quite right unfortunately.

On `master` this is what we have:

Screen Shot 2019-04-12 at 14 07 28

We want to just keep the bottom two styles, but rename them to be 'leading' and 'trailing'. This PR does this:

Screen Shot 2019-04-12 at 14 07 58

So essentially we need to remove 'leading' and 'trailing' in terms of style, and then rename centerLeading and centerTrailing after that.

The best way to do it to avoid confusion might be two use two commits (but in one PR) - one commit to remove leading and trailing entirely, leaving just centerLeading and centerTrailing. Then a second commit to rename centerLeading and centerTrailing to leading and trailing.

PR #111, opened 2019-04-24
[NO-JIRA]: Updated the theming properties for overlay

ojcurtis:
Updated the theming properties for overlay to allow border… radius numbers for when custom button radius is used.

This change allows the theming of border radius for buttons to be implemented on iOS 

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,1 +1,5 @@
 # Unreleased
+
+**Updated:**
+- react-native-bpk-component-touchable-overlay
+  - Updated to support custom number for the `borderRadius` field to now allow `sm`, `lg`, `pill` or `30`
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,4 +2,4 @@
 
 **Updated:**
 - react-native-bpk-component-touchable-overlay
-  - Updated to support custom number for the `borderRadius` field to now allow `sm`, `lg`, `pill` or `30`
+    - `borderRadius` prop now accepts numbers in addition to the existing `sm`, `lg` and `pill` values. E.g. `<BpkTouchableOverlay borderRadius={30} ... />`. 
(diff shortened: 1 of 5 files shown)


UNRELEASED.md
```diff
@@ -1 +1,5 @@
 # Unreleased
+
+**Updated:**
+- react-native-bpk-component-touchable-overlay
+  - Updated to support custom number for the `borderRadius` field to now allow `sm`, `lg`, `pill` or `30`
```
shaundon: I think we can be clearer to end users about what the change actually does and how to use it - what do you reckon of something like this?

```suggestion
    - `borderRadius` prop now accepts numbers in addition to the existing `sm`, `lg` and `pill` values. E.g. `<BpkTouchableOverlay borderRadius={30} ... />`. 
```

packages/react-native-bpk-component-touchable-overlay/stories.js
```diff
@@ -66,4 +66,40 @@ storiesOf('react-native-bpk-component-touchable-overlay', module)
     >
       {viewContent}
     </BpkTouchableOverlay>
+  ))
+  .add('Examples', () => (
```
shaundon:
```suggestion
  .add('All borderRadius values', () => (
```

Less ambiguous when scrolling through the long list

packages/react-native-bpk-component-touchable-overlay/stories.js
```diff
@@ -66,4 +66,40 @@ storiesOf('react-native-bpk-component-touchable-overlay', module)
...
+    <View>
+      <BpkText>Example with Border Radius sm</BpkText>
```
shaundon:
```suggestion
      <BpkText>With borderRadius="sm"</BpkText>
```

..and so on for the below examples, just to make it really clear what we're doing in each one.

PR #120, opened 2019-05-01
[BPK-2411] Add theming support to button border radius

shaundon:
Simulator Screen Shot - iPhone 8 - 2019-05-01 at 15 32 07
Screenshot_1556721113


packages/react-native-bpk-component-nudger/src/BpkNudger-test.common.js
```diff
@@ -93,52 +92,6 @@ const commonTests = () => {
...
-              decreaseButtonLabel="Decrease"
-              increaseButtonLabel="Increase"
-              onChange={() => null}
-            />
-          </BpkThemeProvider>,
-        )
-        .toJSON();
-      expect(tree).toMatchSnapshot();
-      expect(consoleErrorFn.mock.calls.length).toBe(1);
-    });
```
shaundon: These tests broke, and when I went to fix them I realised that they're redundant because they're testing theming on BpkButton, so I don't think that's something BpkNudger should worry about (the theming is a side effect of using button).

packages/react-native-bpk-component-button/src/BpkBorderedButton.android.js
```diff
@@ -58,6 +55,7 @@ const styles = StyleSheet.create({
 export type Props = {
   backgroundColor: string,
   borderColor: string,
+  borderRadius: number,
```
ojcurtis: Could this also be a string based on the types 'sm', 'lg', 'pill'? Therefore flow would give an error?

packages/react-native-bpk-component-button/src/BpkBorderedButton.ios.js
```diff
@@ -61,6 +60,7 @@ const styles = StyleSheet.create({
 export type Props = {
   backgroundColor: string,
   borderColor: string,
+  borderRadius: number,
```
ojcurtis: Could this also be a string based on the types 'sm', 'lg', 'pill'? Therefore flow would give an error?

packages/react-native-bpk-component-button/src/BpkBorderedButton.android.js
shaundon: `BpkTouchableOverlay` accepts a string or number, but `BpkButton` only ever uses numbers for it so it's still valid.

packages/react-native-bpk-theming/src/types.js
```diff
@@ -19,5 +19,40 @@
...
+  navigationBarStatusBarStyle?: string,
+  navigationBarShadowColor?: string,
+  navigationBarPrimaryColor?: string,
+  progressFillBackgroundColor?: string,
+  progressTrackBackgroundColor?: string,
+  starColor?: string,
+  starFilledColor?: string,
+  textFontFamily?: string,
+  flatListSelectedItemColor?: string,
+  textInputFocusedColor?: string,
```
tiagoengel: This is great 👍

UNRELEASED.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+**Added:**
```
tiagoengel: Just a reminder this should be breaking if we do not change the rule to disable theming if props are missing

shaundon:
Here's an example of not supplying the border radius prop to the featured one - it still gets themed but border radius is left alone and no prop type error shows up

Simulator Screen Shot - iPhone 8 - 2019-05-06 at 10 21 55

PR #121, opened 2019-05-03
[BPK-2441] Make iOS buttons consistent

georgegillams:
Align RN iOS button border design to that on iOS Native

## before: 
Screenshot 2019-05-03 at 15 56 45 (2)

## after:
after

### DIFF
--- a/packages/react-native-bpk-component-button/src/BpkBorderedButton.ios.js
+++ b/packages/react-native-bpk-component-button/src/BpkBorderedButton.ios.js
@@ -33,11 +33,13 @@ import { type IconType } from './common-types';
 type ViewProps = ElementProps<typeof View>;
 type ViewStyleProp = $PropertyType<ViewProps, 'style'>;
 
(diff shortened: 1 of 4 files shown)


packages/react-native-bpk-component-button/src/BpkBorderedButton.ios.js
```diff
@@ -108,6 +109,10 @@ const BpkBorderedButton = (props: Props) => {
 
   return (
     <BpkTouchableOverlay
+      overlayStyle={{
+        borderColor: colorGray100,
+        borderWidth: buttonBorderWidth,
+      }}
```
shaundon: Can we include this in the `styles` object for consistency?

georgegillams: Well we could, but `borderColor` is provided as a prop so we would have to set that anyway.

shaundon: Fair enough

PR #125, opened 2019-05-08
[No JIRA] Fix border radius

shaundon:

### DIFF
--- a/packages/react-native-bpk-theming/src/types.js
+++ b/packages/react-native-bpk-theming/src/types.js
@@ -19,7 +19,7 @@
 /* @flow */
 
 export type Theme = {
-  buttonBorderRadius?: Number,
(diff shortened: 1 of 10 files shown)


packages/react-native-bpk-theming/src/types.js
```diff
@@ -19,22 +19,19 @@
 /* @flow */
 
 export type Theme = {
+  buttonBorderRadius?: Number,
```
georgegillams: Should this have changed from `number` to `Number`?

shaundon: Good catch, it's my bloody editor that keeps suggesting it and sneaks it in without me meaning to.

shaundon: Fixed now

PR #133, opened 2019-05-22
[BPK-2496]: Updating gray usage in RN for components beginning with s to p

ojcurtis:

### DIFF
--- a/packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.android.js.snap
+++ b/packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.android.js.snap
@@ -1,101 +1,428 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Android withDivider should render correctly 1`] = `
-<WithDivider
-  mainStyle={null}
-  stub={
-    <withTheme(BpkText)>
-      Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
-    </withTheme(BpkText)>
+<View
+  style={
+    Array [
+      Object {
+        "backgroundColor": "rgb(255, 255, 255)",
+        "borderColor": "rgb(230, 228, 235)",
+        "borderRadius": 4,
+        "borderWidth": 1,
+      },
+      Array [
+        Object {
+          "flexDirection": "row",
+        },
+      ],
+    ]
   }
-  stubStyle={null}
-  theme={null}
-  vertical={false}
 >
-  <withTheme(BpkText)>
-    Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
-  </withTheme(BpkText)>
-</WithDivider>
+  <View
+    style={
+      Array [
+        Object {
+          "flex": 1,
+        },
+      ]
+    }
+  >
+    <Text
+      style={
+        Array [
+          Object {
+            "color": "rgb(82, 76, 97)",
+            "fontFamily": "System",
+            "fontSize": 16,
+            "fontWeight": "400",
+            "letterSpacing": 0.12,
+          },
+          Object {},
+        ]
+      }
+    >
+      Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
+    </Text>
+  </View>
+  <View
+    style={
+      Array [
+        Object {
+          "backgroundColor": "rgb(230, 228, 235)",
+          "flexDirection": "column",
+          "width": 1,
+        },
+      ]
+    }
+  />
+  <View
+    style={
+      Array [
+        Object {
+          "flex": 1,
+        },
+      ]
+    }
+  >
+    <Text
+      style={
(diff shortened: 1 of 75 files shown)


packages/react-native-bpk-component-section-list/src/BpkSectionListHeader.js
```diff
@@ -58,22 +59,49 @@ const styles = StyleSheet.create({
...
+  const textStyles = [styles.text];
+  if (theme) {
+    if (Platform.OS === 'ios') {
+      outerStyles.push({
+        backgroundColor: theme.colorGray50 ? theme.colorGray50 : colorGray50,
+      });
+      textStyles.push({
+        color: theme.colorGray700 ? theme.colorGray700 : colorGray700,
+      });
+    } else if (Platform.OS === 'android') {
```
shaundon: You can just use `else` here as RN only has the two platforms.

packages/react-native-bpk-component-select/package.json
```diff
@@ -20,5 +20,8 @@
     "react-native-bpk-component-text": "^4.1.6",
     "react-native-bpk-component-touchable-native-feedback": "^1.1.54",
     "react-native-bpk-component-touchable-overlay": "^1.2.3"
+  },
+  "devDependencies": {
+    "react-native-bpk-theming": "^1.2.0"
```
shaundon: Should be a dependency

packages/react-native-bpk-component-select/stories.js
```diff
@@ -99,4 +101,36 @@ storiesOf('react-native-bpk-component-select', module)
         validationMessage="A value must be selected to continue"
       />
     </View>
+  ))
+  .add('docs:with-theme', () => (
```
shaundon: Same comment as before about naming 😇

packages/react-native-bpk-component-spinner/src/BpkSpinner.js
```diff
@@ -79,9 +89,9 @@ const BpkSpinner = (props: Props) => {
 
   // Validate that spinner is themeable and correct theming attribute has been
   // supplied. If not, disable theming.
-  if (themeAttributes && spinnerType !== 'primary') {
-    themeAttributes = null;
-  }
+  // if (themeAttributes && spinnerType !== 'primary') {
+  //   themeAttributes = null;
+  // }
```
shaundon: Should probably remove this altogether if it's commented out, or if you're leaving it in because we intend it to be added back in the future just add a quick 'TODO uncomment when grey theming no longer needed'.

packages/react-native-bpk-theming/src/types.js
```diff
@@ -52,4 +52,10 @@ export type Theme = {
   textFontFamily?: string,
   flatListSelectedItemColor?: string,
   textInputFocusedColor?: string,
+  colorGray50?: string,
+  colorGray100?: string,
+  colorGray300?: string,
+  colorGray500?: string,
+  colorGray700?: string,
+  colorGray900?: string,
```
shaundon: 🙌 Nice!

packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.android.js.snap
```diff
@@ -1,287 +1,101 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Android withDivider should render correctly 1`] = `
-<BpkPanel
```
shaundon: It feels odd that all this has been removed, it might be a side effect of something else

packages/react-native-bpk-component-phone-input/src/__snapshots__/BpkDialingCodeList-test.android.js.snap
```diff
@@ -103,7 +103,7 @@ exports[`Android BpkDialingCodeList should render correctly 1`] = `
                   "width": 24,
                 },
                 Array [
-                  null,
+                  undefined,
```
shaundon: These look a bit dodgy - we should find out why `undefined` was added and if it's intentional

packages/react-native-bpk-component-picker/src/__snapshots__/BpkPicker-test.android.js.snap
```diff
@@ -209,11 +204,6 @@ exports[`Android BpkPicker should render correctly with a selected value 1`] = `
     style={
       Object {
         "backgroundColor": "rgba(37, 32, 51, 0.8)",
-        "height": "100%",
-        "left": 0,
-        "position": "absolute",
-        "top": 0,
-        "width": "100%",
```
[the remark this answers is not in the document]
ojcurtis: Found the issue with this, it was how the styles are being applied now but resolved with the correct code change

packages/react-native-bpk-component-phone-input/src/__snapshots__/BpkDialingCodeList-test.android.js.snap
ojcurtis: It seems this has come in since we have wrapped the `BpkFlag` with the `withTheme` - if we change `withTheme(BpkFlag)` to `BpkFlag` then it changes back to `null`.

packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.android.js.snap
ojcurtis: This looks like it is since we added the `withTheme` wrapper around the `WithDivider` component so it now exports the divider rather than it outputting a panel

georgegillams: Yeah @ojcurtis I think you're right. I got round this by using the regular renderer (instead of shallow), so now they look more like the did before 👍

PR #153, opened 2019-06-21
[BPK-2426] Add elevation to primary and featured buttons on Android

shaundon:
* Added elevation to primary and featured buttons. Agreed with Will that secondary and destructive ones won't get elevation as [the Material spec](https://material.io/design/components/buttons.html) only specifies elevation for non-outlined buttons.
* Fixed a bug where buttons that didn't contain icons didn't have the right height. Now, all Android buttons are 36dp which matches Sketch. This was a subtle bug - the difference is less than 1px, but it was much easier to see with the themed buttons that use a different font.


ios/Podfile.lock
```diff
@@ -13,10 +13,10 @@ PODS:
   - glog (0.3.5)
   - React (0.59.0):
     - React/Core (= 0.59.0)
-  - react-native-bpk-component-calendar (0.3.14):
+  - react-native-bpk-component-calendar (1.0.0):
     - Backpack (~> 10.0)
     - React
-  - react-native-bpk-component-dialog (0.1.0):
+  - react-native-bpk-component-dialog (1.0.0):
```
shaundon: This along with lockfiles have changed because they should have been done on `master` already.

packages/react-native-bpk-component-button/src/BpkButtonInner.js
```diff
@@ -21,15 +21,17 @@
...
-import { spacingSm, spacingLg } from 'bpk-tokens/tokens/base.react.native';
+import { spacingSm, fontSizeLg } from 'bpk-tokens/tokens/base.react.native';
 
 import BpkButtonIcon from './BpkButtonIcon';
 import { type IconType } from './common-types';
 
 const styles = StyleSheet.create({
+  text: {
+    lineHeight: fontSizeLg,
+  },
```
shaundon: This is what fixes the bug where buttons without icons weren't 36dp.

packages/react-native-bpk-component-button/src/BpkButtonInner.js
```diff
@@ -38,6 +40,7 @@ const styles = StyleSheet.create({
   iconOnly: {
     marginStart: 0,
     marginEnd: 0,
+    ...(Platform.OS === 'android' ? { lineHeight: fontSizeLg } : {}),
```
shaundon: Moved from above, as we only want this applied when `iconOnly` is true.

PR #160, opened 2019-07-01
[BPK-2605] Icon only buttons always apply circular radius regardless of theming

ojcurtis:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+- react-native-bpk-component-button
+  - `iconOnly` buttons now apply `boarderRadius` theming when applied, to remain pill shaped.
+
 ## How to write a good changelog entry
 
 1. Add 'Breaking', 'Added' or 'Fixed' in bold depending on if the change will be major, minor or patch according to [semver](semver.org).
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -4,7 +4,7 @@
 
 **Fixed:**
 - react-native-bpk-component-button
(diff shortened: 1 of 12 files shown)


UNRELEASED.md
```diff
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+- react-native-bpk-component-button
+  - `iconOnly` buttons no longer apply `boarderRadius` from theming. The iconOnly buttons now remain circular regardless of radius.
```
shaundon: I think this can be more elegantly worded for consumers, how about:

```suggestion
  - `iconOnly` buttons no longer apply `borderRadius` from theming and are always circular.
```

ios/native.xcodeproj/project.pbxproj
```diff
@@ -320,8 +320,10 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
 				en,
 				Base,
+				"he-IL",
```
shaundon: Not sure why this changed (cc @georgegillams)

UNRELEASED.md
georgegillams:
```suggestion
  - `iconOnly` buttons no longer apply `boarderRadius` from theming. `iconOnly` buttons now remain circular regardless of theming radius.
```

shaundon:
The PR title should be a bit more descriptive too - bear in mind we should be able to find this in six months' time and know exactly what happened.

ios/native.xcodeproj/project.pbxproj
georgegillams: Looks fine to me 👍

PR #171, opened 2019-07-16
Backpack RN Dialog events update fix

felixarpa:
RN `BpkDialog` was failing in iOS when the actions were updated. After **updating**, The event emitter kept the old actions and tried to execute them. These actions were null and the app crashed.

With this fix, the all old actions are removed from the event emitter before updating them.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -31,3 +31,5 @@ See [`CHANGELOG.md`](CHANGELOG.md) for real-world examples of good changelog ent
 - `bpk-component-horcrux`:
   - Fixed issue where `BpkHorcrux` would occasionally possess the living.
 
+- `bpk-component-dialog`:
+  - Fixed issue where `BpkDialog` failed in iOS when updating the actions.
--- a/packages/react-native-bpk-component-dialog/src/NativeDialog.ios.js
+++ b/packages/react-native-bpk-component-dialog/src/NativeDialog.ios.js
@@ -59,7 +59,7 @@ class BpkDialog extends Component<Props> {
     this.eventSubscriptions = [];
   }
 
-  componentDidMount() {
+  setUpEvents = () => {
     if (BpkDialogEventEmitter) {
       const { actions, scrimAction } = this.props;
       this.eventSubscriptions.push(
@@ -82,6 +82,16 @@ class BpkDialog extends Component<Props> {
         );
       }
     }
+  };
+
+  componentDidMount() {
+    this.setUpEvents();
+  }
+
+  componentDidUpdate() {
+    this.eventSubscriptions.forEach(subscriber => subscriber.remove());
+    this.eventSubscriptions = [];
+    this.setUpEvents();
   }
 
   componentWillUnmount() {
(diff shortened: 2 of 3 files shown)


UNRELEASED.md
```diff
@@ -31,3 +31,5 @@ See [`CHANGELOG.md`](CHANGELOG.md) for real-world examples of good changelog ent
 - `bpk-component-horcrux`:
   - Fixed issue where `BpkHorcrux` would occasionally possess the living.
 
+- `bpk-component-dialog`:
```
k0nserv: Can you move these to under the `Unreleased` header please? This the example section

packages/react-native-bpk-component-dialog/src/NativeDialog.ios.js
```diff
@@ -26,7 +26,7 @@ import {
   processColor,
 } from 'react-native';
 import type EmitterSubscription from 'react-native/Libraries/vendor/emitter/EmitterSubscription';
-import isNil from 'lodash/isNil';
+import _ from 'lodash';
```
k0nserv: Should we really be importing all of `lodash` like this over specific imports for tree shaking?

packages/react-native-bpk-component-dialog/src/NativeDialog.ios.js
```diff
@@ -47,6 +47,11 @@ export type State = {
   ...$Exact<CommonProps>,
 };
 
+const isArrayEqual = (a: Array<any>, b: Array<any>) =>
+  _(a)
+    .differenceWith(b, _.isEqual)
```
k0nserv: This compares functions with strict equality `===`  which means if you are creating the function in render e.g.

```javascript
<BpkDialog
  {...otherProps}
  actions={[
    {
      text: "Accept",
      type: BUTTON_TYPE.primary,
      callback: () => console.log("Tapped Accept")
    }
  ]}
/>;
```

It will unsub on every render

packages/react-native-bpk-component-dialog/stories.js
```diff
@@ -60,12 +60,18 @@ const disabledScrimAction = {
   callback: () => {},
 };
 
-type DialogState = {
+type Props = {
   ...$Exact<BpkDialogProps>,
+  updatesState: boolean,
```
k0nserv: Can this be more descriptive than `updatesState`?

packages/react-native-bpk-component-dialog/src/NativeDialog.ios.js
felixarpa: Ok. Then I will have to create a custom method to compare actions 🙂

felixarpa: There's not a good way to compare to functions, the best way to ensure they are equals is doing `.toString()` and then compare them.

shaundon:
Nothing has happened on here for a while - is it still being worked on?

felixarpa:
The work has been stopped for a while, I am gonna close it and reopen it again in the future.

PR #177, opened 2019-07-31
[BPK-2749] Bump RN to 0.60.4

tiagoengel:
TODO:

- [x] Fix Linear Gradient
- [x] Update to official backpack-android version
- [x] Test all components
- [x] Migrate iOS
- [x] Fix badge example
- [x] Update accessibility props

The change in `BpkSwitch` is to change the track colour when selected. This has finally been fixed in the new version 🙌 

Screenshot_1564739660

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


package.json
```diff
@@ -109,16 +104,16 @@
...
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.15",
     "markdown-spellcheck": "^1.3.1",
     "metro-react-native-babel-preset": "^0.55.0",
     "prompt": "^1.0.0",
     "prop-types": "^15.5.10",
     "react": "16.8.3",
     "react-dom": "16.8.3",
-    "react-native": "0.59.0",
+    "react-native": "^0.60.4",
```
jmikulka: @tiagoengel Could you please pin this to a specific version: `0.60.4`?

package.json
```diff
@@ -109,16 +104,16 @@
...
     "react-native-linear-gradient": "^2.5.3",
-    "react-test-renderer": "16.8.3",
+    "react-test-renderer": "^16.8.6",
```
jmikulka: Same here.

package.json
```diff
@@ -109,16 +104,16 @@
     "jest": "^24.1.0",
     "lerna": "^3.10.7",
     "lint-staged": "^9.0.0",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.15",
     "markdown-spellcheck": "^1.3.1",
     "metro-react-native-babel-preset": "^0.55.0",
     "prompt": "^1.0.0",
     "prop-types": "^15.5.10",
     "react": "16.8.3",
```
jmikulka: You also need to update `react` and `react-dom` dependencies.

UNRELEASED.md
```diff
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Breaking**:
+
+- Bump react-native to `0.60.4`
```
k0nserv: Is this breaking for all our packages?

package.json
```diff
@@ -109,16 +104,16 @@
     "jest": "^24.1.0",
     "lerna": "^3.10.7",
     "lint-staged": "^9.0.0",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.15",
```
jmikulka: Do you need to bump `lodash` version as part of this upgrade?

ios/Podfile
```diff
@@ -1,39 +1,44 @@
...
+  pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.podspec'
+  pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
 
   # RN Bridging Pods
   pod 'Backpack', '~> 12.0.0'
   pod 'react-native-bpk-component-calendar', path: '../packages/react-native-bpk-component-calendar'
   pod 'react-native-bpk-component-dialog', path: '../packages/react-native-bpk-component-dialog'
 
   # Third party pods
-  pod 'BVLinearGradient', :path => '../node_modules/react-native-linear-gradient'
```
jmikulka: Have you replaced `react-native-linear-gradient` with any alternative solution? If not, what's the reason for its removal?

k0nserv: What happened with `BVLinearGradient`?

android/build.gradle
```diff
@@ -51,12 +55,15 @@ allprojects {
 }
 
 ext {
-    backpackVersion     = "8.0.0"
+    backpackVersion     = "11.0.2"
```
jmikulka: The latest and used below `backpack-android` version is `11.0.3`.

package.json
tiagoengel: Yes, there are packages using `17.11`, `17.14` and `17.15`. I updated and "deduped" it

ios/Podfile
tiagoengel: Is autolinked now, we don't need to import it manually anymore. The same thing is available for Android but it's not working so we still do it manually.

UNRELEASED.md
tiagoengel: Yes because some a11y pros where removed. It will still work but the props are now ignored, so I believe we should mark is as breaking for all packages

ios/Podfile
tiagoengel: It's autolinked now, we don't need to import it manually anymore. The same thing is available for Android but it's not working so we still do it manually.

package.json
tiagoengel: good catch, done!

package.json
tiagoengel: done

package.json
tiagoengel: done

android/build.gradle
tiagoengel: good catch, updated.

packages/react-native-bpk-component-switch/src/BpkSwitch.js
```diff
@@ -56,7 +57,7 @@ const getColors = (themeAttributes: ?Object, value: ?boolean): Object => {
       thumbColor: value ? primaryColor : colorGray50,
       trackColor: {
         false: secondaryColor,
-        true: secondaryColor,
+        true: setOpacity(primaryColor, 0.32), // Taken from here https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/color/MaterialColors.java#L42
```
shaundon: 🎉

PR #188, opened 2019-08-14
[No JIRA] Remove unused deps and decision about them

shaundon:
Greenkeeper sent us a PR to upgrade Acorn (#184). According to our decisions folder, we needed this to make eslint work for some unknown reason. I remember this from when we upgraded Storybook.

I tried removing it and it all works fine now. I tested this by removing and re-cloning this repo at this branch, then `npm install`, `npm test` and `npm start` all work fine, plus I can run the Storybook no problem.

Closes #184.


decisions/dependencies.md
```diff
@@ -1,11 +0,0 @@
-# Dependencies
-
-The following dependencies aren't actively used by this repo but we need them for some other dependencies to work:
-
-* `acorn`
-* `util`
```
shaundon: - We actually are using `util` so this shouldn't be listed here.
- We're not using `corejs` anywhere and it's not listed as a dep.

PR #197, opened 2019-08-22
[No JIRA] Add some extra info about `titleProps`

shaundon:


packages/react-native-bpk-component-flat-list/README.md
```diff
@@ -87,6 +87,10 @@ Inherits all props from React Native's [FlatList](https://facebook.github.io/rea
 | selected           | bool                                  | false    | false         |
 | titleProps         | object                                | false    | {}            |
 
+#### `titleProps`
+
+`titleProps` is passed down to the `BpkText` used for the title. It accepts anything that React Native's [`Text` component](https://facebook.github.io/react-native/docs/text.html#props) does.
```
georgegillams: Why don't we link to BpkText props instead of ReactNative?

UNRELEASED.md
```diff
@@ -1,6 +1,15 @@
...
 
 > Place your changes below this line.
+
+- react-native-bpk-component-flat-list:
+- react-native-bpk-component-section-list:
+  - New `titleProps` prop for passing props through to the underlying text instance.
+
+> Place your changes above this line.
+
+###################################################################
```
georgegillams: What are those!

packages/react-native-bpk-component-flat-list/README.md
shaundon: Because the RN one has an exhaustive list of all props and the BpkText one doesn't. So they'd end up on the RN one eventually anyway.

UNRELEASED.md
shaundon: People keep putting them underneath so I added a massive line as a visual indicator

georgegillams: Seems reasonable

packages/react-native-bpk-component-flat-list/README.md
georgegillams: Cool

UNRELEASED.md
```diff
@@ -1,6 +1,15 @@
 # Unreleased
 
 > Place your changes below this line.
```
ojcurtis: Should we add a line here like the message below

shaundon: Nah I don't think it's needed, the line was just to stop people adding things to the bottom of the file

PR #207, opened 2019-08-29
[BPK-2877] Add destructive badge theming support

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Breaking:**
+
+- react-native-bpk-component-badge:
+  - New required `badgeDestructiveBackgroundColor` theme prop.
+
 > Place your changes above this line.
 
 ###################################################################
--- a/packages/react-native-bpk-component-badge/src/theming.js
+++ b/packages/react-native-bpk-component-badge/src/theming.js
@@ -1,0 +1,26 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { makeThemePropType } from 'react-native-bpk-theming';
+
+const REQUIRED_THEME_ATTRIBUTES = ['badgeDestructiveBackgroundColor'];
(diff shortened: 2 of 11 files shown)


UNRELEASED.md
```diff
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Breaking:**
```
georgegillams: Not sure if this makes it breaking or added...

ojcurtis: It will be added as it didn't exist previously 😄

shaundon: But it's required, which means it'll break badge theming for any existing consumers of it. 

They'll need manually add this theme prop for their code to keep working, which makes this breaking. It's ok to do breaking changes, especially as we only have one consumer of this library and we can manage the upgrade for them.

georgegillams:
> Why is this only for destructive? Shouldn't it be for the yellow and green badges at least too?

Yeah but small PR's are better
Also I have no idea why the tests throw exception now. I'm sure it's obvious but I just can't see it!

UNRELEASED.md
tiagoengel: Unrelated but I think we should relax to whole thing to warning instead of errors (not defining all theme props)

shaundon: It's a prop type warning so it won't actually cause any errors in prod, only dev. But it will disable theming in components that don't provide all the necessary attributes, which I still think is important because we do generally want to ensure that all required props are provided right? Otherwise we run the risk of people seeing partially-themed components which will look bad.

shaundon:
> Yeah but small PR's are better

Surely it's better for consumers to make one breaking PR than three breaking PRs in quick succession? Small PRs generally are better, yes, but this is taking it to an extreme

UNRELEASED.md
ojcurtis: Yeah if not all theme props are provided for a component it will not allow partial theming - it was a change we did a few months ago to prevent what Shaun said about partial themed components

tiagoengel: Right, yeah having partner branding in mind this makes sense

georgegillams:
> > Yeah but small PR's are better
> 
> Surely it's better for consumers to make one breaking PR than three breaking PRs in quick succession? Small PRs generally are better, yes, but this is taking it to an extreme

We can still release once with several breaking changes though...

shaundon:
> We can still release once with several breaking changes though...

That is true, but I would still question the usefulness of doing three different PRs and going through the whole review, CI etc cycle for three changes that are very, very similar. It makes sense when the changes are different but this feels like a waste of effort

packages/react-native-bpk-component-badge/src/theming.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// If theming is ever expanded to support other types, this should be changed
+// to something akin to BpkButton's theming functions.
```
georgegillams: Remove this comment about Buttons 🤦‍♂️

PR #218, opened 2019-09-11
[BPK-2831]: Removed gray theming

ojcurtis:
[BPK-2831]: Deprecated grayForTheme utility
[BPK-2831]: Removed gray theming from components p to t

### DIFF
--- a/packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.ios.js.snap
+++ b/packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.ios.js.snap
@@ -27,6 +27,9 @@ exports[`iOS withDivider should render correctly 1`] = `
         Object {
           "flex": 1,
         },
+        Object {
+          "padding": 16,
+        },
       ]
     }
   >
@@ -63,6 +66,9 @@ exports[`iOS withDivider should render correctly 1`] = `
       Array [
         Object {
           "flex": 1,
+        },
+        Object {
+          "padding": 16,
         },
       ]
     }
@@ -115,6 +121,9 @@ exports[`iOS withDivider should render correctly with arbitrary props 1`] = `
         Object {
           "flex": 1,
         },
+        Object {
+          "padding": 16,
+        },
       ]
     }
   >
@@ -151,6 +160,9 @@ exports[`iOS withDivider should render correctly with arbitrary props 1`] = `
       Array [
         Object {
           "flex": 1,
+        },
+        Object {
+          "padding": 16,
         },
       ]
     }
@@ -203,6 +215,9 @@ exports[`iOS withDivider should render correctly with custom main style 1`] = `
           "flex": 1,
         },
         Object {
+          "padding": 16,
+        },
+        Object {
           "flex": 2,
         },
       ]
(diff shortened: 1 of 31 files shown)


packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.ios.js.snap
```diff
@@ -329,6 +350,9 @@ exports[`iOS withDivider should render correctly with custom stub style 1`] = `
         Object {
           "flex": 1,
         },
+        Object {
+          "padding": 16,
+        },
```
shaundon: Feels suspect this extra padding has appeared.. does the component look ok?

packages/react-native-bpk-component-phone-input/stories.js
```diff
@@ -224,23 +222,15 @@ class FullyIntegrated extends React.Component<
...
 
-storiesOf('react-native-bpk-component-phone-input/Integrated', module)
-  .add('Full component example', () => (
-    <FullyIntegrated initiallySelectedId="CA" codes={codes} />
-  ))
-  .add('Full component example with theme', () => (
-    <BpkThemeProvider theme={themeAttributes}>
-      <FullyIntegrated initiallySelectedId="CA" codes={codes} />
-    </BpkThemeProvider>
-  ));
```
shaundon: Think we still want these themed examples - it still does theming of other colours, just not grey

packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.ios.js.snap
ojcurtis: This might be due to a change with how the component processes it e.g. may have removed an array and changed it to pass through differently - component looked ok but will check it

packages/react-native-bpk-component-phone-input/stories.js
ojcurtis: This component doesn't seem to be using other theming so can't see anything changing

packages/react-native-bpk-theming/src/util.js
```diff
@@ -108,9 +108,9 @@ export const getThemeAttributes = (
 };
 
 export const grayForTheme = (theme: ?Object, colorName: string) => {
-  if (theme && theme[colorName]) {
-    return theme[colorName];
-  }
+  console.warn(
+    'This theming utility has been deprecated and marked for removal. Please update usages of this method to use the Backpack gray pallete',
```
shaundon:
```suggestion
    'This theming utility has been deprecated and marked for removal. Please update usages of this method to use the Backpack gray palette',
```

ojcurtis: Thanks! Never get the spelling right 😆

packages/react-native-bpk-component-panel/src/__snapshots__/withDivider-test.ios.js.snap
ojcurtis: So looks like the `bpk-theming` was removing this padding and now we have removed it from `BpkPanel` its brought it back 🤔 Running it now to check what it looks like - might now conflict with a change we made previously to fix RN panel padding

PR #222, opened 2019-09-23
[NO-JIRA] Fix flow types in bpk icon

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/react-native-bpk-component-icon/src/withRtlSupport.js
+++ b/packages/react-native-bpk-component-icon/src/withRtlSupport.js
@@ -34,7 +34,7 @@ const styles = StyleSheet.create({
 });
 
 // Based loosely on https://github.com/acdlite/recompose/blob/master/src/packages/recompose/getDisplayName.js
-const getDisplayName = Component => {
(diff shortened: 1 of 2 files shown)


packages/react-native-bpk-component-icon/src/withRtlSupport.js
```diff
@@ -46,12 +46,10 @@ const getDisplayName = Component => {
   return Component.displayName || Component.name || 'Component';
 };
 
-export type Props = $Exact<BpkIconProps>;
-
 const withRtlSupport = (
-  IconComponent: ComponentType<any>,
-): ComponentType<any> => {
-  const WithRtlSupport = (props: Props) => {
+  IconComponent: typeof BpkIcon,
```
tiagoengel: Let's make this `ComponentType<BpkIconProps>` otherwise this will be a breaking change. Before we were accepting any component with the same prop definitions as BpkIcon and now we only accept BpkIcon.

Also I see no reason to restrict it to bpk icon only, in that case it would probably be better to just add another export with RTL support

PR #249, opened 2019-10-09
[BPK-2962] Bridge rating component

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.md
```diff
@@ -1,6 +1,11 @@
 # Unreleased
 
 > Place your changes below this line.
+
+**Added:**
+
+- Added Android bridge for `react-native-bpk-component-rating`.
```
shaundon: Should this have a more consumer-friendly description? I'm thinking that the bridging aspect should be invisible to consumers so we should just introduce it like any new component.

packages/react-native-bpk-component-rating/README.md
```diff
@@ -0,0 +1,125 @@
...
+    compileSdkVersion   = 28
+    targetSdkVersion    = 28
+    minSdkVersion       = 21
+    buildToolsVersion   = "28.0.3"
+}
+```
+
+If you haven't or are using the pre compiled version bellow, it will use the values shown above.
+
+#### Pre compiled version
```
shaundon:
```suggestion
#### Pre-compiled version
```

packages/react-native-bpk-component-rating/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+Alternatively, the pre compiled version is available on Skyscanner's internal Artifactory. Make sure you have the `infrastructure-maven` registry configured and are logged in, then add the following dependency to your `build.gradle` file:
```
shaundon:
```suggestion
Alternatively, the pre-compiled version is available on Skyscanner's internal Artifactory. Make sure you have the `infrastructure-maven` registry configured and are logged in, then add the following dependency to your `build.gradle` file:
```

packages/react-native-bpk-component-rating/README.md
```diff
@@ -0,0 +1,125 @@
...
+      implementation 'net.skyscanner.backpack:react-native-bpk-component-rating:<version>'
+    }
+```
+
+**Note:** The version should be the same used for the npm package.
+
+
+#### Importing the bridge package
+
+After you have installed the lib, import the `BpkRatingPackage()` in your react application:
```
shaundon:
```suggestion
After you have installed the lib, import the `BpkRatingPackage()` in your React application:
```

packages/react-native-bpk-component-rating/README.md
```diff
@@ -0,0 +1,125 @@
...
+    return Arrays.<ReactPackage>asList(
+            new MainReactPackage(),
+            new BpkRatingPackage()
+    );
+}
+```
+
+### iOS
+
+TODO
```
shaundon:
```suggestion
This isn't yet implemented for iOS.
```

packages/react-native-bpk-component-rating/README.md
```diff
@@ -0,0 +1,125 @@
...
+    />
+  </View>
+)
+```
+
+## Props
+
+| Property           | PropType                                | Required | Default Value |
+| ------------------ | --------------------------------------- | -------- | ------------- |
+| title              | oneOfType(string, arrayOf(string>))     | true     | -             |
```
shaundon:
```suggestion
| title              | oneOfType(string, arrayOf(string))     | true     | -             |
```

packages/react-native-bpk-component-rating/README.md
```diff
@@ -0,0 +1,125 @@
...
+| subtitle           | oneOfType(string, arrayOf(string>))     | true     | -             |
```
shaundon:
```suggestion
| subtitle           | oneOfType(string, arrayOf(string))     | true     | -             |
```

packages/react-native-bpk-component-rating/README.md
```diff
@@ -0,0 +1,125 @@
...
+| value              | object                                  | true     | -             |
+| icon               | oneOfType(string, arrayOf(string>))     | false    | null          |
```
shaundon:
```suggestion
| icon               | oneOfType(string, arrayOf(string))     | false    | null          |
```

packages/react-native-bpk-component-rating/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "peerDependencies": {
+    "react": "^16.8.3",
+    "react-native": ">= 0.59.0"
+  },
+  "dependencies": {
+    "bpk-tokens": "^28.0.1",
+    "prop-types": "^15.5.8"
+  },
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/"
```
shaundon: Is this a new thing? Haven't seen it before

packages/react-native-bpk-component-rating/src/common-types.js
```diff
@@ -0,0 +1,73 @@
...
+
+type StringOrArray = string | Array<string>;
+
+const SIZES = {
+  icon: 'icon',
+  xs: 'xs',
+  sm: 'sm',
+  base: 'base',
+  lg: 'lg',
+};
```
shaundon: Why don't we export this and let consumers use it? If we do, we should update the readme sample code too.

UNRELEASED.md
tiagoengel: Yeah makes sense, I'll update

packages/react-native-bpk-component-rating/package.json
tiagoengel: not sure, I've copied it from another package so I guess is been there for a while

packages/react-native-bpk-component-rating/src/common-types.js
tiagoengel: My issue with exposing those objects and advertising its usage is that it doesn't make much sense now that we use flow. 

When you are typing code and add `size=` flow will auto suggest the list of valid strings to type in, the same way it will show the list of valid options when it shows an error.

Adding this "emun" object only creates indirection, where instead of typing `size="sm"` you have to type `size={SIZES.sm}` with no benefit. The result of a typo using flow will be mostly the same (`sn` is not supported) and without flow I'd argue the result is worse with the enum object, because in this case you will not see an error but instead the default value for the property will be used, because `sn` will be `undefined`.

tiagoengel:
> I didn't look at the sizing code in depth yet but will try and look today - seems good though.
> 
> One thing to maybe think about it making the package private until the iOS one is done. I'm thinking that we might need to drastically change the RN API depending on if iOS still does the same `[low, medium, high]` thing, so this would cover us from any big breaking changes.

Makes sense, I'll make it private for now. I don't expect the api to change too much though as both platforms have this concept of one title for each score but is better be safe than sorry :)

packages/react-native-bpk-component-rating/src/common-types.js
shaundon: Makes sense to me - thanks for the detailed explanation!

PR #257, opened 2019-10-23
[BPK-2711] Rating Bridge

k0nserv:
Some decision I've made in the course of implementing this:

+ I removed the `icon` size because it's not supported on iOS.
+ I made the prop type for `StringOrArray` better by changing `Array<String>` to `[string, string, string]`. This should catch trivial errors earlier.

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests
+ [ ]  Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -1,3 +1,5 @@
+source 'https://cdn.cocoapods.org/'
+
 platform :ios, '11.0'
 use_modular_headers!
 
@@ -29,9 +31,11 @@ target 'Backpack Native' do
   pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec', :modular_headers => false
 
   # RN Bridging Pods
-  pod 'Backpack', '~> 23.0.0'
+  pod 'Backpack', '~> 23.0'
+
   pod 'react-native-bpk-component-calendar', path: '../packages/react-native-bpk-component-calendar', :modular_headers => false
   pod 'react-native-bpk-component-dialog', path: '../packages/react-native-bpk-component-dialog', :modular_headers => false
+  pod 'react-native-bpk-component-rating', path: '../packages/react-native-bpk-component-rating', :modular_headers => false
 
   # Third party pods
   pod 'react-native-maps', path: '../packages/react-native-bpk-component-map/node_modules/react-native-maps', :modular_headers => false
--- a/package.json
+++ b/package.json
@@ -128,5 +128,6 @@
   },
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
-  }
+  },
+  "dependencies": {}
 }
--- a/packages/react-native-bpk-component-rating/src/BpkRating.js
+++ b/packages/react-native-bpk-component-rating/src/BpkRating.js
@@ -18,6 +18,7 @@
 
 /* @flow */
 import React from 'react';
+import { View } from 'react-native';
 
 import { defaultProps, propTypes, type Props } from './common-types';
 import NativeRating from './NativeRating';
@@ -26,19 +27,29 @@ const arrify = (maybeArray: Array<string> | string): Array<string> =>
   Array.isArray(maybeArray) ? maybeArray : [maybeArray];
 
 const BpkRating = (props: Props) => {
-  const { title, subtitle, icon, ...rest } = props;
+  const { title, subtitle, value, size, orientation, ...outerProps } = props;
+
+  // These two casts are required beacuse Flow cannot cast
+  // `[string, string, string]` to `Array<string> | string`.
+  // This is safe because a three value tuple is just a form
+  // of array.
+  const titleValue: Array<string> | string = (title: any);
+  const subtitleValue: Array<string> | string = (title: any);
 
   return (
-    <NativeRating
-      title={arrify(title)}
-      subtitle={arrify(subtitle)}
-      icon={icon ? arrify(icon) : null}
-      {...rest}
-    />
+    <View {...outerProps}>
+      <NativeRating
+        title={arrify(titleValue)}
+        subtitle={arrify(subtitleValue)}
+        size={size}
+        value={value}
+        orientation={orientation}
+      />
+    </View>
   );
 };
 
-BpkRating.propTypes = { ...propTypes };
-BpkRating.defaultProps = { ...defaultProps };
+BpkRating.propTypes = propTypes;
+BpkRating.defaultProps = defaultProps;
 
 export default BpkRating;
--- a/packages/react-native-bpk-component-rating/src/NativeRating.android.js
+++ b/packages/react-native-bpk-component-rating/src/NativeRating.android.js
@@ -24,12 +24,6 @@ import { type NativeProps } from './common-types';
 
 const AndroidBPKRatingView = requireNativeComponent('AndroidBPKRatingView');
 
-const normalizedIconName = icon => `bpk_${icon.replace(/-/g, '_')}`;
-
-const BpkRating = (props: NativeProps) => {
-  const { icon, ...rest } = props;
-  const normalizedIcon = icon ? icon.map(normalizedIconName) : undefined;
-  return <AndroidBPKRatingView {...rest} icon={normalizedIcon} />;
-};
+const BpkRating = (props: NativeProps) => <AndroidBPKRatingView {...props} />;
 
 export default BpkRating;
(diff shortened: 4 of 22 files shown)


packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTBPKRating.h
```diff
@@ -0,0 +1,31 @@
+/*
```
georgegillams: I don't think we're meant to use the `RCT` namespace 🤔

packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTBPKRating.h
```diff
@@ -0,0 +1,31 @@
...
+#import <Backpack/Rating.h>
+
+NS_ASSUME_NONNULL_BEGIN
+@class RCTBridge;
+@interface RCTBPKRating : BPKRating
+
+- (instancetype)initWithBridge:(RCTBridge *)bridge;
+
+@property(nonatomic, copy, nullable) NSArray<NSString *> *rct_title;
+@property(nonatomic, copy, nullable) NSArray<NSString *> *rct_subtitle;
```
georgegillams: Don't we need a property for the rating value too?

georgegillams: Ignore me... I see it!

packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTBPKRatingShadowView.h
```diff
@@ -0,0 +1,16 @@
+//
```
georgegillams: License header

packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTBPKRatingShadowView.m
```diff
@@ -0,0 +1,38 @@
+//
```
georgegillams: License header

packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTConvert+RCTBPKRating.h
```diff
@@ -0,0 +1,19 @@
+//
```
georgegillams: License header

packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTConvert+RCTBPKRating.m
```diff
@@ -0,0 +1,24 @@
+//
```
georgegillams: License header

packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTBPKRating.h
k0nserv: It's `RCTBPK` not `RCT` we use that for all extensions of Backpack with the purpose of supporting RN

georgegillams: 👍

packages/react-native-bpk-component-rating/src/ios/RatingBridge/RCTBPKRating.m
```diff
@@ -0,0 +1,92 @@
...
+        self.bridge = bridge;
+    }
+
+    return self;
+}
+
+- (void)setNeedsLayout {
+    [super setNeedsLayout];
+
+    CGSize size = [self systemLayoutSizeFittingSize:UILayoutFittingCompressedSize];
```
k0nserv: I dunno how thread safe this is? I think you are only meant to call this on a UI thread, but I think `setNeedsLayout` can be called on other threads

ios/Podfile
```diff
@@ -1,3 +1,5 @@
+source 'https://cdn.cocoapods.org/'
```
georgegillams: what does this do?

package.json
```diff
@@ -128,5 +128,6 @@
   },
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
-  }
+  },
+  "dependencies": {}
```
georgegillams: Don't need this

packages/react-native-bpk-component-rating/src/BpkRating.js
```diff
@@ -26,19 +27,36 @@ const arrify = (maybeArray: Array<string> | string): Array<string> =>
   Array.isArray(maybeArray) ? maybeArray : [maybeArray];
 
 const BpkRating = (props: Props) => {
-  const { title, subtitle, icon, ...rest } = props;
+  const {
+    title,
+    subtitle: any,
```
georgegillams: This seems like a confusing name 🤔

ios/Podfile
k0nserv: It switches us to the [faster new specs repo](http://blog.cocoapods.org/CocoaPods-1.7.2/) for CocoaPods. No more need to run `pod repo update` etc.

package.json
k0nserv: It was auto generated by `npm` I think

packages/react-native-bpk-component-rating/src/BpkRating.js
k0nserv: Why?

ios/Podfile
georgegillams: Nice!

package.json
georgegillams: 🤷‍♀️

packages/react-native-bpk-component-rating/src/BpkRating.js
georgegillams: I thought "any" was the name, but now I realise it was a type annotation... 😆

k0nserv: Np, it did remind me to remove it

packages/react-native-bpk-component-rating/src/NativeRating.android.js
```diff
@@ -20,16 +20,11 @@
...
 
-const normalizedIconName = icon => `bpk_${icon.replace(/-/g, '_')}`;
-
-const BpkRating = (props: NativeProps) => {
-  const { icon, ...rest } = props;
-  const normalizedIcon = icon ? icon.map(normalizedIconName) : undefined;
-  return <AndroidBPKRatingView {...rest} icon={normalizedIcon} />;
-};
+const BpkRating = (props: NativeProps) => <AndroidBPKRatingView {...props} />;
+BpkRating.defaultProps = defaultProps;
```
tiagoengel: This shouldn't have any default props, they come from the `BpkRating` component that wraps this one. Also `NativeProps` doesn't define nullable prop that needs to come from default props.

packages/react-native-bpk-component-rating/src/NativeRating.ios.js
```diff
@@ -17,11 +17,15 @@
...
 import { defaultProps, type NativeProps } from './common-types';
 
-// eslint-disable-next-line
-const BpkRating = (props: NativeProps) => null;
+const RCTBPKRating = requireNativeComponent('RCTBPKRating');
+
+const BpkRating = (props: NativeProps) => <RCTBPKRating {...props} />;
 
-BpkRating.defaultProps = { ...defaultProps };
+BpkRating.defaultProps = defaultProps;
```
tiagoengel: Same thing about the default props

packages/react-native-bpk-component-rating/src/NativeRating.android.js
k0nserv: Not including these breaks Flow

k0nserv: image

tiagoengel: What is the error? It wasn't breaking before and you only removed props 🤔

k0nserv: I fixed it by changing `size?` to `size` and `orientation?` to `orientation`. This is correct because default props are resolved before the function is called so by the time it is `size` and `orientation` will definitely be there

tiagoengel: you are right 👍

PR #260, opened 2019-10-30
[BPK-3175] Create appearance package

tiagoengel:
JS side is ready, I'll include the native bridges and stories in a follow up PR

PR to integrate it in our storybook:
- https://github.com/Skyscanner/backpack-react-native/pull/262

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/react-native-bpk-appearance/src/BpkDynamicStyleSheet.js
+++ b/packages/react-native-bpk-appearance/src/BpkDynamicStyleSheet.js
@@ -58,9 +58,11 @@ export type BpkStyleSheetStyle = {|
 |};
 
 export type NamedStyles = { +[key: string]: BpkStyleSheetStyle };
+
+export type BpkDynamicStyleProp<S> = $ObjMap<S, (Object) => any>;
 export type BpkDynamicStyle<S> = {
-  light: $ObjMap<S, (Object) => any>,
-  dark: $ObjMap<S, (Object) => any>,
+  light: BpkDynamicStyleProp<S>,
+  dark: BpkDynamicStyleProp<S>,
 };
 
 const isSemanticColor = (value: any): boolean =>
--- a/packages/react-native-bpk-appearance/src/BpkDynamicStyleSheet.js
+++ b/packages/react-native-bpk-appearance/src/BpkDynamicStyleSheet.js
@@ -1,0 +1,129 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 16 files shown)


packages/react-native-bpk-appearance/stories.js
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+// import React from 'react';
+// import { storiesOf } from '@storybook/react-native';
+
+// import CenterDecorator from '../../storybook/CenterDecorator';
```
georgegillams: Is this meant to be like this?

packages/react-native-bpk-appearance/src/hooks.js
```diff
@@ -0,0 +1,46 @@
...
+
+export function useColorScheme(): ColorSchemeName {
+  return useAppearance().colorScheme || 'light';
+}
+
+export function useDynamicValue<T>(value: DynamicValue<T>): T {
+  return value[useColorScheme()];
+}
+
+export function useDynamicStyleSheet(style: BpkDynamicStyle<*>) {
```
tiagoengel: This type needs to be improved, it does not report that a style key is missing when doing this 
```
const currentStyle = useDynamicStyleSheet(allStyles);
currentStyle.invalidKey
```

We can do that in a new PR, is not critical

packages/react-native-bpk-appearance/stories.js
tiagoengel: I'll add this when I include the native bridges in a follow up PR

packages/react-native-bpk-appearance/src/hooks.js
tiagoengel: Only need a ping pong game to figure this out :D. It's fix now

packages/react-native-bpk-appearance/src/BpkDynamicStyleSheet.js
```diff
@@ -0,0 +1,130 @@
...
+  ...$Exact<ViewStyle>,
+  ...$Exact<ImageStyle>,
+};
+
+type SenanticColor = { light: string, dark: string };
+type ColorValue<+key: $Keys<AllStylesProps>> =
+  | TypeForStyleKey<key>
+  | SenanticColor;
+
+export type BpkStyleSheetStyle = {|
```
tiagoengel: This is extending react native's stylesheet type and adding `{ light: string, dark: string }` as a valid option for color props. The create function will parse those props and split them into two `StyleSheets` (light and dark) and this can then be used in code like this:

```
const style = BpkDynamicStyleSheet.create({
  view: {
    shadowColor: { light: 'l', dark: 'd' },
  }
}):
// This will have two props, light and dark
// style.light => { view: { color: 'l' }}
// style.dark => { view: { color: 'd' }}

const MyView = () => {
   // The hook will fetch the correct version based on the current color scheme
   const currentStyle = useBpkDynamicStyleSheet(style);
   return <View style={currentStyle.view} />
}```

PR #278, opened 2019-11-13
[NO-JIRA] Add tests for dark mode

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/react-native-bpk-component-text/package.json
+++ b/packages/react-native-bpk-component-text/package.json
@@ -20,7 +20,7 @@
     "react-native-bpk-theming": "^1.4.7"
   },
   "devDependencies": {
-    "react-native-bpk-test-utils": "^0.0.0"
+    "react-native-bpk-test-utils": "^0.0.1"
   },
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
--- a/packages/react-native-bpk-component-text/package.json
+++ b/packages/react-native-bpk-component-text/package.json
@@ -19,6 +19,9 @@
     "react-native-bpk-appearance": "^1.1.2",
     "react-native-bpk-theming": "^1.4.7"
   },
+  "devDependencies": {
+    "react-native-bpk-test-utils": "^0.0.0"
+  },
   "publishConfig": {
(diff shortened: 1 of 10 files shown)


shaundon:
Also, is `0.0.0` the correct version to use for a new package? I might be wrong but I thought that caused issues and it needed to be at least `0.0.1`

tiagoengel:
Yes I thought about automatically applying for every test. Is not that simple because you can only add the wrapper inside the `create` function (or `shallow` and `render` from enzyme) so there is no way to wrap the component from the outside. 

The only other solution is to mock the native bridge to return `dark` or to mock the appearance methods to use a specific variation.

Mocking appearance can be done but is very error prone and hard to debug understand. Consider this example where 'react-native-bpk-test-utils` is also globally mocking the appearance methods.

```javascript
import BpkText from `react-native-bpk-component-text`
// This will work to mock the `Subject` bellow but the text component above will be out of sync
import { describeWithColorScheme } from 'react-native-bpk-test-utils`

import Subject from '.'
```

```javascript
import { describeWithColorScheme } from 'react-native-bpk-test-utils`
// This will now fail to mock the `appearance` package we are including another package that requires it afterwards.
import BpkText from `react-native-bpk-component-text`

import Subject from '.'
```

There are more edge cases but in general is very fragile unless we write the mock logic explicitly in each file, which is them very verbose. 

Mocking the native bridge to return `dark` or `light` for the initial value can work but that will require running all tests for both modes, which will create a lot of unnecessary test duplication because a lot tests don't need to run in dark mode.

tiagoengel:
About the version I'm not sure, but `0.0.1` makes more sense, I'll update. This is also not supposed to be published, it's for internal tests

tiagoengel:
@shaundon other than the reasons above I think a more explicit approach like this is better, is not really a lot of overhead and we can control which tests we want to run in dark mode

shaundon:
@tiagoengel Thanks for the detailed explanation, your logic makes sense :)

PR #281, opened 2019-11-14
[BPK-3234] Add dark mode support to icons

tiagoengel:
I'm updating the default icon color from `colorSkyGrayTint01` to `colorSkyGray` because this is what iOS and Android are using.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -22,18 +22,22 @@ import PropTypes from 'prop-types';
...
+} from 'react-native-bpk-appearance';
 
-const styles = StyleSheet.create({
+const dynamicStyles = BpkDynamicStyleSheet.create({
   icon: {
     fontFamily: 'BpkIcon',
     fontSize: spacingLg,
     includeFontPadding: false,
-    color: colorSkyGrayTint01,
+    color: textPrimaryColor,
```
ojcurtis: As this is an icon and not text is `textPrimaryColor` an appropriate value to be using here?

tiagoengel: It is actually a `text` component :).I would say yes because It is expected to have the same colour as text in normal situations.

It is also the token iOS and Android use

PR #288, opened 2019-11-21
Bpk 3232 add dark mode support to flat list

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/react-native-bpk-component-flat-list/src/common-types.js
```diff
@@ -31,14 +34,16 @@ type ViewStyleProp = $PropertyType<ViewProps, 'style'>;
...
+
 export type FlatListItemProps = {
   onPress: () => mixed,
   title: string,
   selected: boolean,
   image: ?FlatListItemImage,
   style: ViewStyleProp,
   theme: ?Theme,
-  titleProps?: $Shape<BpkTextProps>,
+  titleProps?: $Diff<TextConfig, { children: any }>,
```
tiagoengel: $Shape here wasn't right because it does not allow extra props, only the ones defined in `BpkTextProps` which is not what we want. `Config<BpkTextProps, BpkTextDefaultProps>` behave the same was the when you pass it to the component directly, and I'm removing children because it is not required here and has no default.

PR #295, opened 2019-11-25
Disabling font scaling in profile when system font size is increased

murchant:
Updating the BpkText comment so the resizing does of text does not occur when the phones text is increased through accessibility settings. This is currently the behaviour in native screens and this behaviour should be present in react native screens, and therefore over al screens. 

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -53,4 +53,4 @@ See [`CHANGELOG.md`](CHANGELOG.md) for real-world examples of good changelog ent
   - Fixed issue where `BpkHorcrux` would occasionally possess the living.
 
 - react-native-bpk-component-text:
-  - Disabled font scaling for react native in react-native-bpk-component-text.
+  - Disabled font scaling for REACT NATIVE in react-native-bpk-component-text.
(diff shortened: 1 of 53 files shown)


packages/react-native-bpk-component-text/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-native-bpk-component-text",
-  "version": "4.2.1",
+  "version": "4.2.2",
```
tiagoengel: Was this updated manually? If so, can you rever it? This is handled by the release script.

UNRELEASED.md
```diff
@@ -4,6 +4,9 @@
 
 **Added:**
 
+- react-native-bpk-component-text:
```
tiagoengel: Can you please move this to a `**Fixed:**` block?

packages/react-native-bpk-component-text/package.json
murchant: No I updated it, ok sounds good

UNRELEASED.md
murchant: Yes ! will do

UNRELEASED.md
```diff
@@ -51,3 +51,6 @@ See [`CHANGELOG.md`](CHANGELOG.md) for real-world examples of good changelog ent
 
 - react-native-bpk-component-horcrux:
   - Fixed issue where `BpkHorcrux` would occasionally possess the living.
+
```
tiagoengel: Looks good now, but this should be at the top of the file, this part is just examples.

PR #305, opened 2019-11-27
[BPK-3240] Fix progress bar dark mode

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [ ] Tests


packages/react-native-bpk-component-progress/src/BpkProgress.js
```diff
@@ -77,6 +79,12 @@ const BAR_TYPES = {
   bar: 'bar',
 };
 
+const getStyles = (
```
tiagoengel: Can we remove this and use the `unpack` directly inline. That way flow recognises the type correctly and converting to any is not required

packages/react-native-bpk-component-progress/src/BpkProgress.js
```diff
@@ -91,35 +99,42 @@ export type Props = {
...
-    min: PropTypes.number.isRequired,
-    value: PropTypes.number.isRequired,
-    type: PropTypes.oneOf(Object.keys(BAR_TYPES)),
-    style: ViewPropTypes.style,
-    fillStyle: ViewPropTypes.style,
-    theme: themePropType,
-    accessibilityLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.func])
-      .isRequired,
-  };
+  static propTypes = { ...propTypes };
```
tiagoengel: nit: no need to spreed the props here

packages/react-native-bpk-component-progress/src/BpkProgress.js
georgegillams: Good catch 👍

PR #306, opened 2019-11-27
[BPK-3245] Add dark mode for BpkSwitch and fix 'off' colour on iOS

shaundon:
The 'off' colour was supposed to use `skyGreyTint06` but due to the [confusing API](https://facebook.github.io/react-native/docs/switch) for RN's `<Switch />` component it wasn't actually being set.

This fixes that and adds dark mode support.

We use the same primary blue 'on' colour in both light and dark mode, this mirrors how the platform does it and Will has signed off on it.


packages/react-native-bpk-component-switch/src/BpkSwitch.js
```diff
@@ -43,8 +45,11 @@ const getColors = (themeAttributes: ?Object, value: ?boolean): Object => {
   // The color props mean different things based on the platform.
   const colors = Platform.select({
     ios: {
+      ios_backgroundColor: useBpkDynamicValue({
```
tiagoengel: Can we rename this function to useColors? By convention any function that uses a hook should be name `use` as well.

shaundon: Thanks, updated.

PR #307, opened 2019-11-27
[BPK-3242] Add dark mode support to select

tiagoengel:
Also fixed disabled state and added an example for it.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/react-native-bpk-component-select/src/BpkSelect.js
```diff
@@ -159,7 +166,7 @@ const BpkSelect = (props: Props) => {
         {showImage && styledImage}
         {content || label}
         {validityIcon}
-        <SelectIcon />
+        <SelectIcon disabled={disabled} />
```
georgegillams: How come this has changed?

tiagoengel: This was wrong, the icon had no disabled state before. I fixed it as part of this PR

georgegillams: Cool - nice one 😉

PR #311, opened 2019-11-28
[BPK-3309] Fix text input font theming

tiagoengel:
**NOTE** I also fixed the danger rule that validates if an iOS `fontFamily` token is being used in Android snapshots. 

Previously it was using the wrong prop `FONT_FAMILY.VALUE` instead of `FONT_FAMILY.value`. So it was actually checking for `fontFamily` equals to `undefined`. 

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/react-native-bpk-component-text-input/src/__snapshots__/BpkTextInput-RTL-test.js.snap
+++ b/packages/react-native-bpk-component-text-input/src/__snapshots__/BpkTextInput-RTL-test.js.snap
@@ -1292,7 +1292,7 @@ exports[`RTL BpkTextInput dark mode should support font theming 1`] = `
       style={
         Object {
           "color": "rgba(178, 178, 191, 1)",
-          "fontFamily": "roboto",
+          "fontFamily": "relative",
           "fontSize": 16,
           "fontWeight": "400",
           "lineHeight": 19,
@@ -1337,7 +1337,7 @@ exports[`RTL BpkTextInput dark mode should support font theming 1`] = `
               "textAlign": "right",
             },
             Object {
-              "fontFamily": "roboto",
+              "fontFamily": "relative",
             },
           ]
         }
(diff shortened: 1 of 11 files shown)


packages/react-native-bpk-component-text-input/src/__snapshots__/BpkTextInput-RTL-test.js.snap
```diff
@@ -2638,7 +2793,7 @@ exports[`RTL BpkTextInput light mode should support theming 1`] = `
       style={
         Object {
           "color": "rgba(178, 178, 191, 1)",
-          "fontFamily": "System",
+          "fontFamily": undefined,
```
ojcurtis: Is this expected? Shouldn't it say 'relative'

packages/react-native-bpk-component-text-input/src/__snapshots__/BpkTextInput-test.android.js.snap
```diff
@@ -1292,7 +1368,7 @@ exports[`Android BpkTextInput dark mode should support theming 1`] = `
       style={
         Object {
           "color": "rgba(178, 178, 191, 1)",
-          "fontFamily": "sans-serif",
+          "fontFamily": undefined,
```
ojcurtis: Is this expected?

packages/react-native-bpk-component-text-input/src/__snapshots__/BpkTextInput-RTL-test.js.snap
tiagoengel: It should not be set because this test is only providing the required props, let me double check if the end result is what we expect.

tiagoengel: That was wrong, good catch! Because that text was only providing the required theme attribute the `themeAttributes` variable was not null, but `textFontFamily` was. It's fixed now @ojcurtis

PR #338, opened 2019-12-09
HELIOS-1191: Add a suggested into the country code selection

samstradling:
Adds the ability to have custom suggested codes at the top of the list:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/react-native-bpk-component-phone-input/src/common-types.js
+++ b/packages/react-native-bpk-component-phone-input/src/common-types.js
@@ -38,7 +38,7 @@ export type DialingCodeListProps = {
   renderFlag: RenderFlag,
   onItemPress: Code => void,
   selectedId: ?string,
-  suggested?: {
+  suggested: ?{
     ids: Array<string>,
     title: string,
   },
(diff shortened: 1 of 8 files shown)


packages/react-native-bpk-component-phone-input/src/common-types.js
```diff
@@ -62,6 +66,7 @@ export const LIST_COMMON_PROP_TYPES = {
 
 export const LIST_COMMON_DEFAULT_PROPS = {
   selectedId: null,
+  suggested: undefined,
```
georgegillams: Just to be consistent with other default prop values:

```suggestion
  suggested: null,
```

shaundon:
This is great - only thing I'd add is update the screenshots for the docs site to include the suggested section. We can do that though on our end if you prefer.

Having said that, I reckon we should merge your other PR first and then we can update this one, that way it'll work on both Android and iOS

shaundon:
Argh - I was gonna merge it but I can't because of the conflicts, and I can't fix them myself because it's a fork I don't have access to 🙈 Would you mind fixing the merge conflicts? Or give me access to the fork and I can do it if you're busy

PR #344, opened 2019-12-13
[BPK-3076] Bridge calendar disabled dates props

tiagoengel:
The idea behind this is to avoid sending unnecessary data over the bridge, and to go around the fact we cannot send a function as prop and call in the native side. 

So we send `date matchers` that will be used inside the native side to match dates in the `isDateDisabled` function. E.g: `DateMatchers.after(Date.UTC(2019, 11, 16))` will match all dates after `16/12/2019` and disable them, but it means we don't need to send all dates until the end of the calendar to do that.

Here is an example of the behaviour we are trying to support, allowing to select only 30 days from the `departure` date.

Kapture 2019-12-16 at 13 39 32

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/react-native-bpk-component-calendar/src/NativeCalendar.ios.js
```diff
@@ -86,14 +76,6 @@ class BpkCalendar extends Component<Props, {}> {
       ...rest
     } = this.props;
 
-    if (minDate && maxDate) {
```
tiagoengel: this validation is done in the `BpkCalendar.js`

packages/react-native-bpk-component-calendar/src/DateMatchers.js
```diff
@@ -0,0 +1,112 @@
...
+  dates: dates.map(parseDateToNative), // eslint-disable-line prettier/prettier
+});
+
+const range = createMacher('range');
+const after = createMacher('after');
+const before = createMacher('before');
+const any = createMacher('any');
+
+export default {
+  /**
```
ojcurtis: This inline documentation is 💯

PR #345, opened 2019-12-13
[BPK-3283] Fix prop type warnings for theming

shaundon:
Previously, if you made no attempt to theme a component but placed it inside a `BpkThemeProvider` you would see warnings that you hadn't supplied all the required attributes.

For example:

```
<BpkThemeProvider 
    theme={ /* correct theme props for button but nothing for progress bar */}
>
    <BpkButton />      // Gets themed correctly
    <BpkProgressBar /> // Shows warnings because you didn't pass in its theme props
</BpkThemeProvider>
```

This was fine when theming was only used for partner branding, because we never wanted non-themed components inside a `BpkThemeProvider`. However, consumers are now using theming to just theme _some_ components inside a view hierarchy.

This PR loosens the warning, so that it only fires when _some_ theme attributes _but not all_ for a component are supplied. In other words, when you don't pass in _any_ theme props the component needs, there won't be a warning.

If you still pass in a partial theme - you supply _some_ but not _all_ required attributes, a warning will still appear.

Here's an example - I added a `BpkProgressBar` to our storybook but didn't pass in any of its theme props. It appears fine using the default styling, and no warning was shown.

Simulator Screen Shot - iPhone 8 - 2019-12-13 at 19 25 36

I also did a couple of other small housekeeping things in this PR, which I'll comment on inline.

### DIFF
--- a/packages/react-native-bpk-component-progress/README.md
+++ b/packages/react-native-bpk-component-progress/README.md
@@ -28,8 +28,8 @@ export default class App extends Component {
   render() {
     return (
       <View style={styles.container}>
-        <BpkProgress min={0} max={100} value={10} />
-        <BpkProgress min={0} max={100} value={10} type="Bar" />
+        <BpkProgress min={0} max={100} value={10} accessibilityLabel="0 of 100" />
+        <BpkProgress min={0} max={100} value={10} accessibilityLabel="0 of 100" type="Bar" />
       </View>
     );
   }
@@ -46,7 +46,7 @@ export default class App extends Component {
 | value              | number                                | true     | -             |
 | fillStyle          | object                                | false    | -             |
 | style              | object                                | false    | -             |
-| theme              | See [Theme Props](#theme-props) below | false    | -             | 
+| theme              | See [Theme Props](#theme-props) below | false    | -             |
 | type               | oneOf('default', 'bar')               | false    | default       |
 
 
--- a/packages/react-native-bpk-theming/src/util-test.js
+++ b/packages/react-native-bpk-theming/src/util-test.js
@@ -35,6 +35,10 @@ const VALID_THEME = {
   themeAttributeC: true,
   nonRequiredAttribute: 'green',
 };
+// When zero required attributes are passed in, the theme is still valid.
+const VALID_EMPTY_THEME = {
+  nonRequiredAttribute: 'green',
+};
 const INVALID_THEME = {
   themeAttributeA: 'red',
   themeAttributeC: true,
@@ -44,6 +48,10 @@ const INVALID_THEME = {
 describe('isValidTheme', () => {
   it('should validated themes that are valid', () => {
     expect(isValidTheme(REQUIRED_ATTRIBUTES, VALID_THEME)).toBeTruthy();
+  });
+
+  it('should validated themes when none of the required props are passed in', () => {
+    expect(isValidTheme(REQUIRED_ATTRIBUTES, VALID_EMPTY_THEME)).toBeTruthy();
   });
 
   it('should fail to validate invalid themes', () => {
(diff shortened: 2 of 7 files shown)


packages/react-native-bpk-component-progress/README.md
```diff
@@ -28,8 +28,8 @@ export default class App extends Component {
   render() {
     return (
       <View style={styles.container}>
-        <BpkProgress min={0} max={100} value={10} />
-        <BpkProgress min={0} max={100} value={10} type="Bar" />
+        <BpkProgress min={0} max={100} value={10} accessibilityLabel="0 of 100" />
+        <BpkProgress min={0} max={100} value={10} accessibilityLabel="0 of 100" type="Bar" />
```
shaundon: `accessibilityLabel` is a required prop so we should include it in our sample code.

packages/react-native-bpk-theming/src/util-test.js
```diff
@@ -115,18 +123,14 @@ describe('makeThemePropType', () => {
       ).toBeFalsy();
     });
 
-    it('should fail when random props are provided', () => {
+    it('should not fail when random props are provided', () => {
```
shaundon: So, this test failed when I made my change. From what I can tell though, we really don't mind if random props are passed in, so I changed it.

packages/react-native-bpk-theming/stories.js
```diff
@@ -45,15 +44,6 @@ type Theme = {
...
   buttonSecondaryBorderColor: string,
-  primaryColor: string,
-  colorGray50: string,
-  colorSkyGrayTint06: string,
-  colorSkyGrayTint05: string,
-  colorSkyGrayTint04: string,
-  colorSkyGrayTint03: string,
-  colorSkyGrayTint02: string,
-  colorSkyGrayTint01: string,
-  colorSkyGray: string,
```
shaundon: This was all for the brand refresh and it's no longer needed, or even supported? Idk but let's get rid of it

packages/react-native-bpk-theming/stories.js
```diff
@@ -219,7 +97,7 @@ class BpkThemePicker extends Component<{}, State> {
     this.themes = {
       blue: generateThemeAttributes(
         colorSkyBlue,
-        Platform.OS === 'ios' ? colorSkyBlueTint01 : colorSkyBlue,
+        colorSkyBlue,
```
shaundon: This doesn't seem needed

packages/react-native-bpk-theming/src/util-test.js
tiagoengel: 👍

packages/react-native-bpk-theming/src/util.js
```diff
@@ -119,7 +138,9 @@ export const getThemeAttributes = (
     return null;
   }
 
-  const hasAllRequiredAttributes = isValidTheme(requiredAttributes, theme);
+  const hasAllRequiredAttributes =
```
tiagoengel: Why are we not using `isValidTheme` here, it seems to do the same?

shaundon:
> Also, do we a have a test to check that providing just a few required props still fails? e.g if we have 3 required props providing two or one should show the warning.

Not sure - I'll check and add if necessary.

packages/react-native-bpk-theming/src/util.js
shaundon: Because `isValidTheme` returns `true` if zero required props are passed in, whereas in this case we *don't* want to pass a theme down to components if there are zero passed in.

Maybe this is confusing and it should be refactored.. what do you think?

tiagoengel: Got it, makes sense. I guess we can add a comment explaining this as it not that clear

PR #347, opened 2019-12-18
[BPK-3379] Add new colorBuckets prop to BpkCalendar

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/react-native-bpk-component-calendar/README.md
+++ b/packages/react-native-bpk-component-calendar/README.md
@@ -270,6 +270,10 @@ Creates a new color bucket to be used in BpkCalendar.
 
 A color bucket is used to define custom colours for calendar days.
 
+NOTE: Your are responsible for ensuring multiple color buckets don't
+overlap, in case they do the last one applied (last in the list) will
+take precedence.
+
 #### Parameters
 
 -   `color` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** The background color
@@ -283,8 +287,8 @@ A color bucket is used to define custom colours for calendar days.
 ```javascript
 <BpkCalendar
  colorBuckets={[
-   colorBucket(colorPanjin, DateMatchers.range(startOfSumer, endOfSumer)),
-   colorBucket(colorSagano, DateMatchers.after(endOfSumer))
+   colorBucket(colorPanjin, DateMatchers.range(startOfSummer, endOfSummer)),
+   colorBucket(colorSagano, DateMatchers.after(endOfSummer))
  ]}
 />
 ```
--- a/packages/react-native-bpk-component-calendar/README.md
+++ b/packages/react-native-bpk-component-calendar/README.md
@@ -145,6 +145,7 @@ class App extends Component {
 | Property              | PropType               | Required | Default Value          |
 | --------------------- | ---------------------- | -------- | ---------------------- |
 | locale                | string                 | true     | -                      |
+| colorBuckets          | arrayOf(ColorBucket)   | false    | undefined              |
 | disabledDates         | DateMatcher            | false    | null                   |
 | maxDate               | oneOf(Date, number)    | false    | today + 1 year         |
 | minDate               | oneOf(Date, number)    | false    | today                  |
@@ -176,6 +177,9 @@ class App extends Component {
     -   [any](#any)
         -   [Parameters](#parameters-3)
         -   [Examples](#examples-3)
+-   [colorBucket](#colorbucket)
+    -   [Parameters](#parameters-4)
+    -   [Examples](#examples-4)
 
 ### DateMatchers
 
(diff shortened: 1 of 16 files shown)


packages/react-native-bpk-component-calendar/README.md
```diff
@@ -259,3 +263,30 @@ An any matcher will match if the date is equal to any of the dates provided.
...
+-   `textStyle` **TextStyle?** The text style. Valid values are `light` or `dark`.
+       Changes how the text looks based on the background color, where light or dark refers
+       to the background colour. (optional, default `undefined`)
+
+#### Examples
+
+```javascript
+<BpkCalendar
+ colorBuckets={[
+   colorBucket(colorPanjin, DateMatchers.range(startOfSumer, endOfSumer)),
```
shaundon:
```suggestion
   colorBucket(colorPanjin, DateMatchers.range(startOfSummer, endOfSummer)),
```

packages/react-native-bpk-component-calendar/README.md
```diff
@@ -259,3 +263,30 @@ An any matcher will match if the date is equal to any of the dates provided.
...
+   colorBucket(colorSagano, DateMatchers.after(endOfSumer))
```
shaundon:
```suggestion
   colorBucket(colorSagano, DateMatchers.after(endOfSummer))
```

packages/react-native-bpk-component-calendar/README.md
```diff
@@ -259,3 +263,30 @@ An any matcher will match if the date is equal to any of the dates provided.
...
+
+Creates a new color bucket to be used in BpkCalendar.
+
+A color bucket is used to define custom colours for calendar days.
+
+#### Parameters
+
+-   `color` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** The background color
+-   `days` **DateMatcher** The days in this bucket
+-   `textStyle` **TextStyle?** The text style. Valid values are `light` or `dark`.
```
shaundon: Should this be an enum? e.g. `TextStyles.LIGHT`

packages/react-native-bpk-component-calendar/src/colorBucket-test.js
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { processColor } from 'react-native';
```
shaundon: How on Earth did you find this, I couldn't see it documented in RN anywhere, had to look in the source 😂

packages/react-native-bpk-component-calendar/src/common-types.js
```diff
@@ -106,4 +109,5 @@ export const commonDefaultProps = {
   selectionType: SELECTION_TYPES.single,
   style: null,
   disabledDates: null,
+  colorBuckets: undefined,
```
shaundon: Why `undefined` and not `null`?

shaundon:
Runs great locally 😍 

One small thing - the label on the picker says 'disable dates' which isn't right. Can we change it?

Screenshot_1576744560

packages/react-native-bpk-component-calendar/src/common-types.js
tiagoengel: undefined is the same as not providing the prop whereas null will add the prop with a null value. it's manly to not pollute snapshots

packages/react-native-bpk-component-calendar/src/colorBucket-test.js
k0nserv: I think I found it that way too :)

packages/react-native-bpk-component-calendar/README.md
```diff
@@ -259,3 +263,30 @@ An any matcher will match if the date is equal to any of the dates provided.
...
+### colorBucket
+
+Creates a new color bucket to be used in BpkCalendar.
+
+A color bucket is used to define custom colours for calendar days.
+
+#### Parameters
+
+-   `color` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** The background color
+-   `days` **DateMatcher** The days in this bucket
```
k0nserv: What happens if you have overlapping DateMatchers?

tiagoengel: Last one takes precedence, meant to write that down but forgot

packages/react-native-bpk-component-calendar/README.md
tiagoengel: it can but I think is irrelevant if if we use flow and say the type is `light | dark`. I think we had this discussion before, but the object version `TextStyles.LIGHT` has no benefit over strings and is more verbose, specially considering we use flow

packages/react-native-bpk-component-calendar/src/colorBucket-test.js
tiagoengel: Found it by accident at some point, can't remember when now :D

PR #349, opened 2019-12-18
[BPK-2705] Fix flow types in calendar

k0nserv:
The flow types for the external Calendar component did not match the
internal native components.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/react-native-bpk-component-calendar/src/common-types.js
+++ b/packages/react-native-bpk-component-calendar/src/common-types.js
@@ -20,6 +20,7 @@
 
 import { type ElementProps } from 'react';
 import { View, ViewPropTypes } from 'react-native';
+import { type SyntheticEvent } from 'react-native/Libraries/Types/CoreEventTypes';
 import PropTypes from 'prop-types';
 
 import { type DateMatcher } from './DateMatchers';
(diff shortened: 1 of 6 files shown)


packages/react-native-bpk-component-calendar/src/common-types.js
```diff
@@ -34,13 +35,26 @@ export const SELECTION_TYPES = {
 };
 
 export type SelectionType = $Keys<typeof SELECTION_TYPES>;
+export type SelectedDatesChanged = ?(Date[]) => mixed;
+export type NativeEvent = SyntheticEvent<$ReadOnly<{| selectedDates: Date[] |}>,>; // eslint-disable-line prettier/prettier
```
tiagoengel: aren't native dates a number? (timestamp)

k0nserv: Oh yeah you're right

tiagoengel: Also just noticed there is an extra comma at the end.
And you could replace $ReadOnly with `+{| selectedDates: Date[] |}` :)

k0nserv: The extra comma gets added by prettier

k0nserv: I tried replacing it with `+{| selectedDates: number[] |}`, but that causes a syntax error.

PR #350, opened 2019-12-19
[BPK-3324] Add search component for FlatList and SectionList

shaundon:
Adds search to flat list and section list. It's as lightweight as possible, it's basically:

- A component for users to use with FlatList and SectionList's `ListHeaderComponent` prop.
- A component for users to use with FlatList and SectionList's `ListEmptyComponent` prop.
- Associated tests, readmes and screenshots.

Doing the actual filtering is left to userland. I've implemented it in the stories though for an example.

✅ Uses native look on both platforms (approved by Will)
✅ Dark mode support

Kapture 2019-12-19 at 17 28 28


packages/react-native-bpk-component-flat-list/stories.js
```diff
@@ -60,25 +62,28 @@ class StatefulBpkFlatList extends React.Component<
...
+    includeSearch: false,
   };
 
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
     this.itemPressCallbacks = {};
-    this.state = { selectedCountry: 'DJ' };
+    this.state = { selectedCountry: 'DJ', countries: this.getData() };
```
tiagoengel: getData depends on a prop so we shouldn't only fetch it in the constructor. We should memoize the function or use https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

tiagoengel: Sorry, didn't realise this a `story` file 🤦‍♂ . It's fine here.

PR #353, opened 2019-12-20
[BPK-2470] Add outline button type

shaundon:
This took literally five minutes, so glad we did that refactor a few months ago


packages/react-native-bpk-component-button/stories.js
```diff
@@ -38,6 +38,10 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     flexWrap: 'wrap',
     justifyContent: 'flex-start',
+    padding: spacingMd,
+  },
+  btnContainerDark: {
+    backgroundColor: colorAbisko,
```
tiagoengel: nit: what about using black so we keep a dark background in dark mode? 👼

shaundon: But then I'd be using black in light mode too, which is against our brand 😉

shaundon:
Have updated the readme

PR #354, opened 2019-12-20
[BPK-2806] Reinstall storybook server.

tiagoengel:
This reinstall the storybook server and disables the mobile UI. The main motivation is that the UI doesn't support RTL and so testing is very hard. Plus it is a lot more annoying to use.

All addons are also back online 🎉 .

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


storybook/addons/actions.js
```diff
@@ -0,0 +1,72 @@
...
+      const parser = createParser();
+      return Object.entries(a).reduce(parser, {});
+    }),
+]);
+
+/**
+ * Creates an action that can handle cyclic object structures as StoryBook is not able to handle
+ * this currently.
+ *
+ * Note that this might end up removing some props from the ation object so don't relly 100% on
```
shaundon:
```suggestion
 * Note that this might end up removing some props from the action object so don't rely 100% on
```

PR #355, opened 2019-12-23
[BPK-2806] Adjust text RTL behaviour

tiagoengel:
Couple of findings:

**iOS**
Works fine as longs as the language is a right to left language, only using `I18Manager.forceRtl(true)` doesn't work for text if the language itself is not rtl.

**Android**
Only works if the language is rtl and it has specific characters of that language, meaning that setting the language to Arabic and using English text doesn't make the text align to the right side. In this case `textAlign: left` trick fixes it.

I've tested this in the app and haven't noticed any difference other than that RTL works on Android for the situation described above.

It is also not possible to properly test RTL in our current setup because the storybook UI does not support it, that's why I have this PR to reinstall the storybook server https://github.com/Skyscanner/backpack-react-native/pull/354.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


tiagoengel:
Let's merge this only after we finish with dark mode changes to avoid any problems.

shaundon:
> Let's merge this only after we finish with dark mode changes to avoid any problems.

@tiagoengel we're probably done with that now right, can we merge this?

PR #356, opened 2019-12-24
[NO-JIRA] Fix android tests

tiagoengel:
I found out most of our Android tests were not using the correct tokens because even though we mock it in the Android files, it does not work for imports of imports, e.g. If we are testing Badge and Badge is also import Text, the mock will not reflect into the Text component, even if we reset the mocks and dynamically require everything again.

This solution runs the test twice simulating better how the code would actually run, and always resolving `ios` or `android` files.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


tiagoengel:
closed in favour of #361

PR #359, opened 2020-01-07
[No Jira] Fix tests on Android

shaundon:
https://github.com/Skyscanner/backpack-react-native/pull/356 is failing, both locally and on CI. I couldn't figure out why so I decided to start from scratch and recreate it bit by bit until it broke. IT NEVER BROKE. So here it is again, let's see if CI passes I guess


tiagoengel:
closed in favour of #361

PR #361, opened 2020-01-14
[NO-JIRA] Update test config

tiagoengel:
This PR replaces the jest rn preset with the expo jest presets. Those presets allow us to run iOS and Android test as a separated jest project and removes the need to mock react-native and other imports to resolve the Android version.

For now tests that don't have `.android` or `.ios` at the end run as iOS tests, this is how it was before so it should not cause any problems. In the future we can change those to run once for each platform if we see the need.

Not added in this PR but also supported by the new config is naming tests `.native.js` so they run for both iOS and Android. **This means we can get rid of the majority of our `.ios`, `.android` and `.common` files and only use one `.native` file**. Will do this at some point in another PR.

This is how the test output looks like now:
Screenshot 2020-01-14 at 17 50 52

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -67,13 +67,13 @@ PODS:
     - React-cxxreact (= 0.60.4)
     - React-jsi (= 0.60.4)
   - React-jsinspector (0.60.4)
-  - react-native-bpk-component-calendar (4.2.0):
+  - react-native-bpk-component-calendar (4.2.1):
     - Backpack (~> 25.0)
     - React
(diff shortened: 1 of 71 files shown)


tiagoengel:
Not sure the link is necessary because jest-expo is a package and has good docs https://github.com/expo/expo/tree/cee2253dbcac0d382931dfdaa45260d53ad86a78/packages/jest-expo.

We just use it directly.

PR #370, opened 2020-01-24
[BPK-3515] Replace underlying bpk-appearance library

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -3,10 +3,17 @@
 > Place your changes below this line.
 
 <<<<<<< HEAD
+<<<<<<< HEAD
 =======
 **Fixed:**
  - react-native-bpk-appearance:
    - We've changed the underlying library used to enable dark-mode support.
+=======
+**Breaking:**
+
+- react-native-bpk-appearance:
+  - We've changed the underlying library used to enable dark-mode support.
+>>>>>>> Update android
 
 >>>>>>> [BPK-3515] Replace underlying bpk-appearance library
 ## How to write a good changelog entry
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,13 @@
 
 > Place your changes below this line.
 
+<<<<<<< HEAD
+=======
+**Fixed:**
+ - react-native-bpk-appearance:
+   - We've changed the underlying library used to enable dark-mode support.
+
+>>>>>>> [BPK-3515] Replace underlying bpk-appearance library
 ## How to write a good changelog entry
 
 1. Add 'Breaking', 'Added' or 'Fixed' in bold depending on if the change will be major, minor or patch according to [semver](semver.org).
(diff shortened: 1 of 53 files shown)


UNRELEASED.md
```diff
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Breaking:**
+
+- react-native-bpk-appearance:
+  - We've changed the underlying library used to enable dark-mode support.
```
shaundon:
```suggestion
  - Changed the underlying library that's used to enable Dark Mode support.
```

packages/react-native-bpk-appearance/README.md
```diff
@@ -65,18 +65,18 @@ Alternatively, the pre compiled version is available on Skyscanner's internal Ar
 
 #### Importing the bridge package
 
-After you have installed the lib, import the `DialogPackage()` in your react application:
+After you have installed the lib, import the `RNCAppearancePackage()` in your react application:
```
shaundon:
```suggestion
After you have installed the lib, import the `RNCAppearancePackage()` in your React application:
```

packages/react-native-bpk-appearance/package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "bpk-tokens": "^30.1.0",
     "prop-types": "^15.5.8",
-    "react-native-dark-mode": "^0.2.1-rc.0"
+    "react-native-appearance": "github:tiagoengel/react-native-appearance"
```
shaundon: Why a fork? Just out of interest. Also, if we're gonna officially use a fork it should be under the `skyscanner` org.

tiagoengel: It's a fork because the official one doesn't work correctly on android unless you add a bit of code the the native app, which to me seems wrong and I'd like to avoid adding this to our app. I have this PR to change it so it would only be a fork until that is released, or if they refuse merging this for some reason we will need to add this extra code to the app.
https://github.com/expo/react-native-appearance/pull/35

shaundon: Ah ok cool. If, for some reason, we end up maintaining a fork for a long time, we should move it to Skyscanner. But this is all good for now.

tiagoengel: I'll move the fork to `Skyscanner` org :)

georgegillams: Might be worth adding an explanation to the README too so that it doesn't get "cleaned up"

shaundon:
@georgegillams @tiagoengel do you remember what's going on with this PR? Should we rebase and merge, or close?

georgegillams:
I think we realised this created more issues than it solved, so were gonna hold off and see if RN comes up with a proper way of supporting this

PR #378, opened 2020-02-12
[BPK-3410]: Deprecated BpkThemeAttributes component

olliecurtis:
Added a warning message to be thrown for those still using this package to upgrade before we remove permenantly.

### DIFF
--- a/packages/react-native-bpk-theming/src/BpkThemeAttributes.js
+++ b/packages/react-native-bpk-theming/src/BpkThemeAttributes.js
@@ -89,6 +89,11 @@ export type Props = {
  *
  */
 const BpkThemeAttributes = (props: Props) => {
+  // eslint-disable-next-line no-console
+  console.warn(
+    'The BpkThemeAttributes component has been deprecated and marked for removal. Please remove usages of this component to utilise the brand palette',
+  );
+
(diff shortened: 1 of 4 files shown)


packages/react-native-bpk-theming/src/BpkThemeAttributes.js
```diff
@@ -89,6 +89,11 @@ export type Props = {
  *
  */
 const BpkThemeAttributes = (props: Props) => {
+  // eslint-disable-next-line no-console
```
georgegillams: I think we have a helper for this warning, but maybe that was only for deprecated proptypes

georgegillams: I think we have a helper for this warning, but maybe that was only for deprecated proptypes

olliecurtis: Will have a look, for this I just took the same logic from the `util.js` file where `grayForTheme` was deprecated

olliecurtis: Only seems to exist for propTypes in `packages/react-native-bpk-component-text/src/common-types.js` so not even a global function

PR #380, opened 2020-02-13
[BPK-3547] Update bpk tokens

tiagoengel:
This adds tokens with flow typing and fixes a few errors.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/react-native-bpk-appearance/src/hooks-flowtest.js
```diff
@@ -0,0 +1,127 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* eslint-disable backpack/use-tokens */
+/* @flow */
+import { useBpkDynamicValue, useBpkDynamicStyle } from './hooks';
+import { type BpkDynamicValue } from './common-types';
```
tiagoengel: This file is only here to ensure most cases continue to work as we change the flow config, because it is very hard to know if all possible cases are working otherwise

packages/react-native-bpk-component-progress/src/BpkProgress.js
```diff
@@ -22,7 +22,7 @@ import React, { Component, type ElementProps, type Config } from 'react';
 import PropTypes from 'prop-types';
 import { View, Animated, Easing, Platform, ViewPropTypes } from 'react-native';
 import {
-  animmationDurationBase,
+  animationDurationBase,
```
tiagoengel: 🙈

packages/react-native-bpk-appearance/src/hooks-flowtest.js
olliecurtis: Given that most of the lines here are `eslint-disable-line` would it make sense to just do `eslint-disable` for the whole file?

packages/react-native-bpk-appearance/src/common-types.js
```diff
@@ -24,13 +24,13 @@ export type BpkDynamicValue<T> = { light: T, dark: T };
...
 // prettier-ignore
 export type UnpackedBpkDynamicValue<X> = $Call<
-  & ((BpkDynamicValue<string>) => string)
-  & ((BpkDynamicValue<number>) => number)
-  & ((BpkDynamicValue<boolean>) => boolean)
-  & ((BpkDynamicValue<Array<string>>) => Array<string>)
-  & ((BpkDynamicValue<Array<number>>) => Array<number>)
-  & ((BpkDynamicValue<Array<boolean>>) => Array<boolean>)
-  & ((BpkDynamicValue<*>) => any)
+  & ((string | BpkDynamicValue<string>) => string)
```
georgegillams: Surely `BpkDynamicValue<string>` includes `string`?
If not shouldn't we change `BpkDynamicValue` instead of every usage of it?

PR #383, opened 2020-02-13
[BPK-3381] iOS native calendar colouring bridge

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-25 at 07 03 56

### DIFF
--- a/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.h
+++ b/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.h
@@ -20,6 +20,7 @@
 #import <Backpack/SimpleDate.h>
 #import <React/RCTViewManager.h>
 
+@class RCTBPKColorBucket;
 @interface RCTBPKCalendar : BPKCalendar
 
 @property(nonatomic, copy) RCTBubblingEventBlock onDateSelection;
@@ -27,5 +28,6 @@
 
 @property(nonatomic, nullable) NSDate *rct_minDate;
 @property(nonatomic, nullable) NSDate *rct_maxDate;
+@property(nonatomic, copy, nullable) NSArray<RCTBPKColorBucket *> *rct_colorBuckets;
 @property(nonatomic, nonnull) NSArray<NSDate *> *rct_selectedDates;
 @end
--- a/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
+++ b/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
@@ -19,6 +19,8 @@
 #import "RCTBPKCalendar.h"
 
 #import "RCTBPKCalendarDateUtils.h"
+#import "RCTBPKColorBucket.h"
+#import "RCTBPKDateMatcher.h"
 
 NS_ASSUME_NONNULL_BEGIN
 @interface RCTBPKCalendar ()
@@ -125,6 +127,24 @@ NS_ASSUME_NONNULL_BEGIN
     }
 
     [super setSelectedDates:simpleDates];
+}
+
+- (void)setRct_colorBuckets:(NSArray<RCTBPKColorBucket *> *_Nullable)rct_colorBuckets {
+    _rct_colorBuckets = rct_colorBuckets;
+
+    for (NSInteger i = 0; i < rct_colorBuckets.count; i += 1) {
+        RCTBPKColorBucket *bucket = rct_colorBuckets[i];
+
+        NSMutableArray *localDates = [[NSMutableArray alloc] initWithCapacity:bucket.days.dates.count];
+        for (int j = 0; j < bucket.days.dates.count; j += 1) {
+            localDates[j] = [RCTBPKCalendarDateUtils convertDateToLocal:bucket.days.dates[j]
+                                                          localCalendar:self.gregorian
+                                                            utcCalendar:self.utcCalendar];
+        }
+
+        bucket.days.dates = localDates;
+    }
+    [self reloadData];
 }
 
 @end
--- a/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarDateUtils.m
+++ b/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarDateUtils.m
@@ -39,5 +39,15 @@ NS_ASSUME_NONNULL_BEGIN
     return [localCalendar dateFromComponents:utcComponents];
 }
 
++ (BOOL)date:(NSDate *)date isAfterDate:(NSDate *)reference {
+    NSComparisonResult ordering = [reference compare:date];
+    return ordering == NSOrderedAscending;
+}
+
++ (BOOL)date:(NSDate *)date isBeforeDate:(NSDate *)reference {
+    NSComparisonResult ordering = [reference compare:date];
+    return ordering == NSOrderedDescending;
+}
+
 @end
 NS_ASSUME_NONNULL_END
--- a/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
+++ b/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
@@ -20,12 +20,15 @@
 #import <React/RCTBridge.h>
 #import <React/RCTUIManager.h>
 
+#import <Backpack/Color.h>
+
 #import "RCTBPKCalendar.h"
 #import "RCTBPKCalendarDateUtils.h"
+#import "RCTBPKColorBucket.h"
+#import "RCTBPKDateMatcher.h"
 #import "RCTConvert+RCTBPKCalendar.h"
 
 @interface RCTBPKCalendarManager () <BPKCalendarDelegate>
-
 @end
 
 @implementation RCTBPKCalendarManager
@@ -41,6 +44,7 @@ RCT_EXPORT_MODULE()
 
 RCT_REMAP_VIEW_PROPERTY(minDate, rct_minDate, NSDate)
 RCT_REMAP_VIEW_PROPERTY(maxDate, rct_maxDate, NSDate)
+RCT_REMAP_VIEW_PROPERTY(colorBuckets, rct_colorBuckets, RCTBPKColorBucketArray);
 RCT_EXPORT_VIEW_PROPERTY(selectionType, BPKCalendarSelection)
 RCT_EXPORT_VIEW_PROPERTY(locale, NSLocale)
 RCT_REMAP_VIEW_PROPERTY(selectedDates, rct_selectedDates, NSArray<NSDate *> *)
@@ -105,6 +109,93 @@ RCT_EXPORT_METHOD(forceRender : (nonnull NSNumber *)reactTag) {
     }
 
     calendar.onDateSelection(@{@"selectedDates": dateArray});
+}
+
+- (RCTBPKColorBucket *_Nullable)colorBucketForDate:(NSDate *)date
+                                      colorBuckets:(NSArray<RCTBPKColorBucket *> *)colorBuckets {
+    for (RCTBPKColorBucket *bucket in colorBuckets) {
+        switch (bucket.days.matcherType) {
+        case RCTBPKDateMatcherTypeRange:
+            if ([RCTBPKCalendarDateUtils date:date isAfterDate:bucket.days.dates[0]] &&
+                [RCTBPKCalendarDateUtils date:date isBeforeDate:bucket.days.dates[1]]) {
+                return bucket;
+            }
+            break;
+        case RCTBPKDateMatcherTypeBefore:
+            if ([RCTBPKCalendarDateUtils date:date isBeforeDate:bucket.days.dates[0]]) {
+                return bucket;
+            }
+            break;
+        case RCTBPKDateMatcherTypeAfter:
+            if ([RCTBPKCalendarDateUtils date:date isAfterDate:bucket.days.dates[0]]) {
+                return bucket;
+            }
+            return nil;
+        case RCTBPKDateMatcherTypeAny:
+            for (NSDate *bucketDate in bucket.days.dates) {
+                if ([bucketDate isEqualToDate:date]) {
+                    return bucket;
+                }
+            }
+            break;
+        }
+    }
+    return nil;
+}
+
+- (UIColor *)calendar:(BPKCalendar *)calendar fillColorForDate:(NSDate *)date {
+    NSAssert([calendar isKindOfClass:RCTBPKCalendar.class],
+             @"calendar value is not of type RCTBPKCalendar as expected.");
+    if (![calendar isKindOfClass:RCTBPKCalendar.class]) {
+        return BPKColor.clear;
+    }
+
+    RCTBPKCalendar *rctCalendar = (RCTBPKCalendar *)calendar;
+    RCTBPKColorBucket *colorBucket = [self colorBucketForDate:date colorBuckets:rctCalendar.rct_colorBuckets];
+
+    if (colorBucket) {
+        return colorBucket.color;
+    }
+    return BPKColor.clear;
+}
+
+- (UIColor *)calendar:(BPKCalendar *)calendar titleColorForDate:(NSDate *)date {
+    NSAssert([calendar isKindOfClass:RCTBPKCalendar.class],
+             @"calendar value is not of type RCTBPKCalendar as expected.");
+    if (![calendar isKindOfClass:RCTBPKCalendar.class]) {
+        return BPKColor.textPrimaryColor;
+    }
+
+    RCTBPKCalendar *rctCalendar = (RCTBPKCalendar *)calendar;
+    RCTBPKColorBucket *colorBucket = [self colorBucketForDate:date colorBuckets:rctCalendar.rct_colorBuckets];
+
+    if (colorBucket) {
+        switch (colorBucket.textStyle) {
+        case RCTBPKColorBucketTextStyleDefault:
+            return [self.class getTextColorForBackground:colorBucket.color];
+        case RCTBPKColorBucketTextStyleLight:
+            return BPKColor.textPrimaryLightColor;
+        case RCTBPKColorBucketTextStyleDark:
+            return BPKColor.textPrimaryDarkColor;
+        }
+    }
+    return BPKColor.textPrimaryColor;
+}
+
+#pragma mark - Helpers
+
++ (UIColor *)getTextColorForBackground:(UIColor *)backgroundColor {
+    CIColor *ciColor = [[CIColor alloc] initWithColor:backgroundColor];
+
+    // Luminance calculation using the "Relative luminance in colorimetric spaces" algorithm
+    // https://en.wikipedia.org/wiki/Relative_luminance#Relative_luminance_in_colorimetric_spaces
+    double luminance = 0.2126 * ciColor.red + 0.7152 * ciColor.green + 0.0722 * ciColor.blue;
+
+    if (luminance < 0.5) {
+        return BPKColor.textPrimaryDarkColor;
+    } else {
+        return BPKColor.textPrimaryLightColor;
+    }
 }
 
 @end
--- a/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKColorBucket.h
+++ b/packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKColorBucket.h
@@ -1,0 +1,45 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2020 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Backpack/SimpleDate.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RCTBPKDateMatcher;
+@interface RCTBPKColorBucket : NSObject
+
+/**
+ * Enum values for specifying calendar selection type
+ */
+typedef NS_ENUM(NSUInteger, RCTBPKColorBucketTextStyle) {
+    RCTBPKColorBucketTextStyleDefault = 0,
+    RCTBPKColorBucketTextStyleLight = 1,
+    RCTBPKColorBucketTextStyleDark = 2,
+};
+
+@property(nonatomic, readonly, strong) UIColor *color;
+@property(nonatomic, readonly, assign) RCTBPKColorBucketTextStyle textStyle;
+@property(nonatomic, readonly, strong, nullable) RCTBPKDateMatcher *days;
+
+- (instancetype)initWithColor:(UIColor *)color
+                    textStyle:(RCTBPKColorBucketTextStyle)textStyle
+                         days:(RCTBPKDateMatcher *)days;
+
+@end
+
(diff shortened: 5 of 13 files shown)


packages/react-native-bpk-component-calendar/stories.js
```diff
@@ -424,7 +424,11 @@ storiesOf('react-native-bpk-component-calendar', module)
             minDate={minDate}
             maxDate={maxDate}
             colorBuckets={[
-              colorBucket(colorSagano, dateMatchers[selectedValue].descriptor),
+              colorBucket(
+                colorSagano,
+                dateMatchers[selectedValue].descriptor,
+                'light',
```
tiagoengel: Why adding the default value?

georgegillams: The default value is undefined, by the style should be light on `colorSagano`

tiagoengel: There is a bit of mismatch in how Android and iOS deal with this then, Android will use the best value based on the bucket colour when it's undefined.

georgegillams: Oh right ok - we should do the same for each platform I guess

georgegillams: What algorithm do you use to determine luminance of the background color?

tiagoengel:
RN LGTM!

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -20,23 +20,28 @@
...
 #import "RCTBPKCalendar.h"
 #import "RCTBPKCalendarDateUtils.h"
 #import "RCTConvert+RCTBPKCalendar.h"
+#import "RCTBPKColorBucket.h"
+#import "RCTBPKDateMatcher.h"
 
 @interface RCTBPKCalendarManager () <BPKCalendarDelegate>
-
+@property(nonatomic, strong) NSArray<RCTBPKColorBucket *> *colorBuckets;
+@property(nonatomic, strong) RCTBPKCalendar *calendar;
```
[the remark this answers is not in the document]
k0nserv: The data should live either in the view instance itself or in the shadow view. For example [`RCTTextShadowView`](https://github.com/facebook/react-native/blob/86cad7d69099baaaeae10ac7c0912a023101f257/Libraries/Text/Text/RCTTextShadowView.h#L14) keeps data within it that is later used as part of layout. Since our case doesn't affect the layout or Yoga I think storing the buckets inside `RCTBPKCalendar` is the right choice here

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +114,83 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
     calendar.onDateSelection(@{@"selectedDates": dateArray});
 }
 
+- (RCTBPKColorBucket *_Nullable) colourBuckedForDate:(NSDate *)date {
```
k0nserv: `colour` -> `color` 🇺🇸

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +114,83 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+    return BPKColor.textPrimaryColor;
+}
+
+#pragma Helpers
+
+- (UIColor *)getTextColorForBackground:(UIColor *)backgroundColor {
+    CIColor *ciColor = [[CIColor alloc] initWithColor:backgroundColor];
+
+    // Luminance calculation using the "Relative luminance in colorimetric spaces" algorithm
+    double luminance = 0.2126 * ciColor.red + 0.7152 * ciColor.green + 0.0722*ciColor.blue;
```
k0nserv: Neat 👍 Can we include a Wikipedia link in the comments too?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +114,83 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+
+    if (luminance < 0.5) {
+        return BPKColor.textPrimaryDarkColor;
+    } else {
+        return BPKColor.textPrimaryLightColor;
+    }
+}
+
+- (BOOL)date:(NSDate *)date isAfterDate:(NSDate *)reference {
```
k0nserv: This doesn't need to be an instance method

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +114,83 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+    return [[date laterDate:reference] isEqualToDate:date];
+}
+
+- (BOOL)date:(NSDate *)date isBeforeDate:(NSDate *)reference {
```
k0nserv: Same as above

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKColorBucket.h
```diff
@@ -0,0 +1,43 @@
...
+ * Enum values for specifying calendar selection type
+ */
+typedef NS_ENUM(NSUInteger, RCTBPKColorBucketTextStyle) {
+    RCTBPKColorBucketTextStyleDefault = 0,
+    RCTBPKColorBucketTextStyleLight = 1,
+    RCTBPKColorBucketTextStyleDark = 2,
+};
+
+@property(nonatomic) UIColor *color;
+@property(nonatomic) RCTBPKColorBucketTextStyle textStyle;
```
k0nserv: Can we be explicit about `assing` here?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKColorBucket.h
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Enum values for specifying calendar selection type
+ */
+typedef NS_ENUM(NSUInteger, RCTBPKColorBucketTextStyle) {
+    RCTBPKColorBucketTextStyleDefault = 0,
+    RCTBPKColorBucketTextStyleLight = 1,
+    RCTBPKColorBucketTextStyleDark = 2,
+};
+
+@property(nonatomic) UIColor *color;
```
k0nserv: Can we be explicit about `strong` here?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKColorBucket.h
```diff
@@ -0,0 +1,43 @@
...
+@property(nonatomic) RCTBPKColorBucketTextStyle textStyle;
+@property(nonatomic, nullable) RCTBPKDateMatcher *days;
```
k0nserv: Can we be explicit about `strong` here?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKDateMatcher.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+typedef NS_ENUM(NSUInteger, RCTBPKDateMatcherType) {
+    RCTBPKDateMatcherTypeRange = 0,
+    RCTBPKDateMatcherTypeAfter = 1,
+    RCTBPKDateMatcherTypeBefore = 2,
+    RCTBPKDateMatcherTypeAny = 3,
+};
+
+@property(nonatomic) RCTBPKDateMatcherType matcherType;
+@property(nonatomic, nullable) NSArray<NSDate *> *dates;
```
k0nserv: Keep in mind that we use different time zone conventions on the RN side and inside the calendar so this needs some converting otherwise I think it will not be accurate. See the `setRct_*` methods on `RCTBPKCalendar`

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKColorBucket.h
georgegillams: Assing? 🐴

k0nserv: `assign`^

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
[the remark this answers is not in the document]
georgegillams: This will require a change to this delegate, right?
Because if we don't have a reference to the Calendar we would have to be told by the delegate method what calendar it relates to...
And then cast the calendar to a RCTBPKCalendar to get the colorBuckets?

https://github.com/Skyscanner/backpack-ios/blob/master/Backpack/Calendar/Classes/BPKCalendar.h#L76-L81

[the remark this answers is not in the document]
k0nserv: Yes 👍 All delegate methods should have `calendar:` as the first argument by convention anyway. We can keep those old methods and deprecate them to avoid a breakign change

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.h
```diff
@@ -20,12 +20,14 @@
 #import <Backpack/SimpleDate.h>
 #import <React/RCTViewManager.h>
 
+@class "RCTBPKColorBucket.h";
```
k0nserv: This isn't valid syntax AFAIK. It should be

```objective-c
@class RCTBPKColorBucket;
```

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.h
```diff
@@ -20,12 +20,14 @@
...
 @interface RCTBPKCalendar : BPKCalendar
 
 @property(nonatomic, copy) RCTBubblingEventBlock onDateSelection;
 @property(nonatomic, strong, readonly) NSCalendar *utcCalendar;
 
 @property(nonatomic, nullable) NSDate *rct_minDate;
 @property(nonatomic, nullable) NSDate *rct_maxDate;
+@property(nonatomic, nullable) NSArray<RCTBPKColorBucket *> *rct_colorBuckets;
```
k0nserv: Can we specify `strong` on this and the other properties that don't specify it or `copy`?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
```diff
@@ -127,5 +129,27 @@ - (void)setRct_selectedDates:(NSArray<NSDate *> *)rct_selectedDates {
     [super setSelectedDates:simpleDates];
 }
 
+- (void)setRct_colorBuckets:(NSArray<RCTBPKColorBucket *> *_Nullable)rct_colorBuckets {
+    if(_rct_colorBuckets == rct_colorBuckets) {
```
k0nserv: This is doing a pointer comparison, is that intended? Will it ever be the same?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
```diff
@@ -127,5 +129,27 @@ - (void)setRct_selectedDates:(NSArray<NSDate *> *)rct_selectedDates {
...
+        return;
+    }
+
+    _rct_colorBuckets = rct_colorBuckets;
+
+    for (int i = 0; i < rct_colorBuckets.count ; i+= 1) {
+        RCTBPKColorBucket *bucket = rct_colorBuckets[i];
+
+        NSMutableArray *localDates = [[NSMutableArray alloc]initWithCapacity:bucket.days.dates.count];
```
k0nserv: Missing a space here before `init...`

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
```diff
@@ -127,5 +129,27 @@ - (void)setRct_selectedDates:(NSArray<NSDate *> *)rct_selectedDates {
...
+        for (int j = 0; j < bucket.days.dates.count ; j+= 1) {
+            localDates[j] = [RCTBPKCalendarDateUtils convertDateToLocal:bucket.days.dates[j]
```
k0nserv: Can we cleanup the method call syntax here, maybe align the colons or whatever Xcode does by default?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
```diff
@@ -127,5 +129,27 @@ - (void)setRct_selectedDates:(NSArray<NSDate *> *)rct_selectedDates {
...
 }
 
+- (void)setRct_colorBuckets:(NSArray<RCTBPKColorBucket *> *_Nullable)rct_colorBuckets {
+    if(_rct_colorBuckets == rct_colorBuckets) {
+        return;
+    }
+
+    _rct_colorBuckets = rct_colorBuckets;
+
+    for (int i = 0; i < rct_colorBuckets.count ; i+= 1) {
```
k0nserv: Why is there an extra space after count but not after `0`?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarDateUtils.m
```diff
@@ -39,5 +39,13 @@ + (NSDate *)convertDateToLocal:(NSDate *)utcDate
     return [localCalendar dateFromComponents:utcComponents];
 }
 
++ (BOOL)date:(NSDate *)date isAfterDate:(NSDate *)reference {
+    return [[date laterDate:reference] isEqualToDate:date];
```
k0nserv: Wouldn't it be cleaner to use `compare:` instead?

```objective-c
    NSComparisonResult ordering = [date compare:reference];

    return ordering == NSOrderedDescending || ordering == NSOrderedSame
```

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarDateUtils.m
```diff
@@ -39,5 +39,13 @@ + (NSDate *)convertDateToLocal:(NSDate *)utcDate
...
+}
+
++ (BOOL)date:(NSDate *)date isBeforeDate:(NSDate *)reference {
+    return [[date earlierDate:reference] isEqualToDate:date];
```
k0nserv: Same as above

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +109,82 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
     calendar.onDateSelection(@{@"selectedDates": dateArray});
 }
 
+- (RCTBPKColorBucket *_Nullable) colorBuckedForDate:(NSDate *)date colorBuckets:(NSArray<RCTBPKColorBucket *> *)colorBuckets {
```
k0nserv: Can we remove the space before `colorBuckedForDate:`?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +109,82 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+                }
+                break;
+        }
+    }
+    return nil;
+}
+
+- (UIColor *)calendar:(BPKCalendar *)calendar fillColorForDate:(NSDate *)date {
+    RCTBPKCalendar *rctCalendar = (RCTBPKCalendar *)calendar;
+    if(rctCalendar != nil) {
```
k0nserv: Why would this ever happen? By definition if we are in a delegate method of a calendar the `calendar` parameter will never be `nil`

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +109,82 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+        RCTBPKColorBucket *colorBucket = [self colorBuckedForDate:date colorBuckets:rctCalendar.rct_colorBuckets];
+
+        if(colorBucket) {
```
k0nserv: Spacing

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
k0nserv: In fact it's not marked as `_Nullable` so it violates our own API

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +109,82 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+            return colorBucket.color;
+        }
+    }
+    return BPKColor.clear;
+}
+
+- (UIColor *)calendar:(BPKCalendar *)calendar titleColorForDate:(NSDate *)date {
+    RCTBPKCalendar *rctCalendar = (RCTBPKCalendar *)calendar;
+    if(rctCalendar != nil) {
```
k0nserv: Same as above

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +109,82 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+                    return BPKColor.textPrimaryDarkColor;
+            }
+        }
+    }
+    return BPKColor.textPrimaryColor;
+}
+
+#pragma Helpers
+
+- (UIColor *)getTextColorForBackground:(UIColor *)backgroundColor {
```
k0nserv: nitpic: This doesn't need to be an instance method

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -105,4 +109,82 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+                    return BPKColor.textPrimaryLightColor;
+                case RCTBPKColorBucketTextStyleDark:
+                    return BPKColor.textPrimaryDarkColor;
+            }
+        }
+    }
+    return BPKColor.textPrimaryColor;
+}
+
+#pragma Helpers
```
k0nserv: Should be

```objective-c
#pragma mark - Helpers
```

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTConvert+RCTBPKCalendar.m
```diff
@@ -27,6 +30,38 @@ @implementation RCTConvert (RCTBPKCalendar)
...
+                   @"any": @(RCTBPKDateMatcherTypeAny)
+               }),
+               RCTBPKColorBucketTextStyleDefault, integerValue)
+
 RCT_ARRAY_CONVERTER(NSDate)
+RCT_ARRAY_CONVERTER(RCTBPKColorBucket)
+
++ (RCTBPKDateMatcher *)RCTBPKDateMatcher:(id)json {
+    RCTBPKDateMatcherType matcherType = [RCTConvert RCTBPKDateMatcherType:json[@"type"]];
+    NSArray<NSDate *> *dates = [RCTConvert NSDateArray:json[@"dates"]];
```
k0nserv: I don't think this is still handling the timezone issues correctly

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
georgegillams: Use an assert to check that the calendar is a `RCTBPKCalendar.class`

georgegillams:
Need to test this in extreme time-zone cases 👍

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTConvert+RCTBPKCalendar.m
georgegillams: That's handled in `RCTBPKCalendar setRct_colorBuckets`

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
```diff
@@ -107,4 +111,83 @@ - (void)calendar:(RCTBPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKS
...
+
+        if (colorBucket) {
+            return colorBucket.color;
+        }
+    }
+    return BPKColor.clear;
+}
+
+- (UIColor *)calendar:(BPKCalendar *)calendar titleColorForDate:(NSDate *)date {
+    NSAssert([calendar isKindOfClass:RCTBPKCalendar.class],
```
k0nserv: All asserts need to be coupled with an if using the reversed condition for production since asserts do nothing in production builds. Without it this code would crash at runtime in production builds due to `colorBuckedForDate:colorBuckets:` being unrecgonized

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKColorBucket.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Enum values for specifying calendar selection type
+ */
+typedef NS_ENUM(NSUInteger, RCTBPKColorBucketTextStyle) {
+    RCTBPKColorBucketTextStyleDefault = 0,
+    RCTBPKColorBucketTextStyleLight = 1,
+    RCTBPKColorBucketTextStyleDark = 2,
+};
+
+@property(nonatomic, strong) UIColor *color;
```
k0nserv: Can these properties be `readonly`?

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendar.m
```diff
@@ -127,5 +129,23 @@ - (void)setRct_selectedDates:(NSArray<NSDate *> *)rct_selectedDates {
     [super setSelectedDates:simpleDates];
 }
 
+- (void)setRct_colorBuckets:(NSArray<RCTBPKColorBucket *> *_Nullable)rct_colorBuckets {
+    _rct_colorBuckets = rct_colorBuckets;
+
+    for (int i = 0; i < rct_colorBuckets.count; i += 1) {
```
k0nserv: We should use `NSUInteger` over `int` in Objective-C

packages/react-native-bpk-component-calendar/src/ios/CalendarBridge/RCTBPKCalendarManager.m
georgegillams: Done

PR #384, opened 2020-02-14
[BPK-3567]: Upgrade RN to 0.61.5

olliecurtis:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/react-native-bpk-theming/src/BpkThemeProvider.js
```diff
@@ -22,6 +22,16 @@ import createWithTheme from './create-with-theme';
 
 const BpkThemeProvider = ThemeProvider;
 
+// TODO: Upgrade theming package and remove this: https://gojira.skyscanner.net/browse/BPK-3639
```
tiagoengel: I added this because this causes a lot of warnings making it really hard to see anything important. The same thing happens with `@skyscanner/react-native-transitiongroup` and `recyclerlistview` but those are only used in the carousel and not really a big issue. I have also added cards to upgrade those

olliecurtis: Should we be having a link to our internal tools here in the source?

.flowconfig
```diff
@@ -23,6 +23,9 @@ node_modules/warning/.*
 ; Flow doesn't support platforms
 .*/Libraries/Utilities/HMRLoadingView.js
 
+; Flow doesn't support platforms
+.*/Libraries/Utilities/LoadingView.js
```
olliecurtis: Ah that's how you get around react itself throwing errors!

packages/react-native-bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -101,7 +101,8 @@ class AnimateAndFade extends Component<Props> {
     this.measure(captureHeightAndAnimate);
   }
 
-  componentWillReceiveProps(nextProps: Props) {
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps: Props) {
```
olliecurtis: Should we be fixing this now rather than waiting for the point it is removed?

packages/react-native-bpk-component-banner-alert/src/__snapshots__/BpkBannerAlert-test.android.js.snap
```diff
@@ -317,7 +317,7 @@ exports[`Android BpkBannerAlert dark mode should render correctly with dismissab
             accessibilityRole="button"
             accessibilityStates={Array []}
             accessible={true}
-            clickable={true}
+            focusable={true}
```
olliecurtis: Is this change intentional?

packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.android.js.snap
```diff
@@ -19,7 +19,7 @@ exports[`Android API levels Android API<21 should render without ripple 1`] = `
     accessibilityRole="button"
     accessibilityStates={Array []}
     accessible={true}
-    clickable={true}
+    focusable={true}
```
olliecurtis: Same as previous comment

packages/react-native-bpk-component-card/src/__snapshots__/BpkCard-test.android.js.snap
```diff
@@ -16,7 +16,7 @@ exports[`Android BpkCard dark mode should render correctly 1`] = `
     accessibilityLabel="Example Card"
     accessibilityRole="button"
     accessible={true}
-    clickable={true}
+    focusable={true}
```
olliecurtis: Same as previous comment

packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -104,7 +104,8 @@ class BpkHorizontalNav extends React.Component<
     };
   }
 
-  componentWillReceiveProps(nextProps) {
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
olliecurtis: Should we fix this now to use the correct methods?

packages/react-native-bpk-component-navigation-bar/src/isIphoneX.js
```diff
@@ -23,8 +23,12 @@ export default (() => {
   if (Platform.OS === 'android') {
     return false;
   }
+  const windowDimensions = DeviceInfo.getConstants().Dimensions.window;
+  if (!windowDimensions) {
```
olliecurtis: Could you clarify the logic and what is happening here?

packages/react-native-bpk-theming/src/BpkThemeProvider.js
```diff
@@ -22,6 +22,16 @@ import createWithTheme from './create-with-theme';
 
 const BpkThemeProvider = ThemeProvider;
 
+// TODO (BPK-3639): Upgrade theming package and remove this
+const cwm = ThemeProvider.prototype.componentWillMount;
```
olliecurtis: Could you clarify what this logic is doing here?

.flowconfig
tiagoengel: Yeah, in this case is flow's inability to understand platform imports, `.ios` and `.android`, so we need to ignore this file.

packages/react-native-bpk-component-banner-alert/src/AnimateAndFade.js
tiagoengel: We should fix deprecations as soon as possible, this will be an error in the next release so is better to update now while we still have time to catch any problems. Plus this causes a very big an annoying warning.

packages/react-native-bpk-component-banner-alert/src/__snapshots__/BpkBannerAlert-test.android.js.snap
tiagoengel: Yes, this is a change in how RN's `TouchableNativeFeedback` is rendered. It doesn't affect the actual usage of the component.

packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.android.js.snap
tiagoengel: yes

packages/react-native-bpk-component-banner-alert/src/AnimateAndFade.js
olliecurtis: Can we/should we be doing this in this PR? Or do we have an action to do that later?

packages/react-native-bpk-component-navigation-bar/src/isIphoneX.js
tiagoengel: The interface for this has changed (https://github.com/facebook/react-native/pull/24878/files) and `window` is an `iOS` only property, and it can be null here because this file is not platform specific. So we first need to check if it is present before we use it. The end result should be the same as before

olliecurtis: Ok sounds great 😄

packages/react-native-bpk-theming/src/BpkThemeProvider.js
tiagoengel: I'm just renaming `componentWillMount` and `componentWillReceiveProps` to `UNSAFE_componentWillMount` and `UNSAFE_componentWillReceiveProps` to remove the deprecation warnings. Once we update the theming package we can remove this, but because we don't expose it directly to consumers we can do this little hack to remove the warning for now and avoid polluting the app with even more warnings.

olliecurtis: I am wondering if we should be hacking this instead of just moving to the correct replacements and fixes for this?

packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
tiagoengel: We should fix this as soon as we can, but I don't think we should as part of this update because is more changes to introduce errors in an already big change. So I'd rather do the RN update first and then look at removing these methods.

olliecurtis: Ok sounds good - we should make an issue for this 😄

packages/react-native-bpk-theming/src/BpkThemeProvider.js
tiagoengel: The only thing we can do to remove this is upgrading the theming lib, and it's a breaking change, so I'd rather not do that with the RN upgrade.

packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
tiagoengel: We have one in the backlog already to remove deprecated methods :)

PR #387, opened 2020-02-18
[BPK-3636] Enable relative for Android app

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/storybook/RelativeFontProvider.js
+++ b/storybook/RelativeFontProvider.js
@@ -1,0 +1,46 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2020 Skyscanner Ltd
+ *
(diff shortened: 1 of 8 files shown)


storybook/RelativeFontProvider.js
```diff
@@ -0,0 +1,46 @@
+/*
```
olliecurtis: Do we need a test file with tests for this?

tiagoengel: Not really, it's part of the example app/storybook, we don't test or publish any of this. If it's not working will be pretty clear anyway

olliecurtis: In which case all is good from me!

PR #400, opened 2020-03-06
[BPK-3718] Improve release process

tiagoengel:
This removes the `release-it` lib because of a few issues we where having with it:

- It tries to pushes local branches and fails to release.
- We couldn't find a way to make it run only the npm commands in the `dist` folder and all else in root.
- We were using very little of it to be worth the trouble.

I also increased the `dryRun` coverage so now all process can be tested without publishing anything. This is more or less how the output looks:

Screenshot 2020-03-06 at 09 35 37

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -38,4 +38,9 @@ mkdir -p "$DEST"
 
 cp -r lib/* lib/.npmignore "$DEST"
 
+# We link instead of copying because we want the version update to be applied
+# to the lib folder as well.
+#
+ln -s -f $PWD/lib/package.json "$DEST"
+
 cp README.md CHANGELOG.md LICENSE dist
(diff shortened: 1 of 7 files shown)


scripts/build.sh
```diff
@@ -38,4 +38,9 @@ mkdir -p "$DEST"
 
 cp -r lib/* lib/.npmignore "$DEST"
 
+# We link instead of copying because we want the version update to be applied
+# to the lib folder as well.
+#
+ln -s -f $PWD/lib/package.json "$DEST"
```
georgegillams: btw you can do `-sf` instead of `-s -f`. But it doesn't matter which you use!

shaundon:
Tbh I didn't run this locally, I just reviewed the diff. I figure it'll be tested thoroughly as you do releases

PR #401, opened 2020-03-06
[NO-JIRA] Remove mp4 from npm package

tiagoengel:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/lib/.npmignore
+++ b/lib/.npmignore
@@ -6,3 +6,4 @@ screenshots
 *-test.js
 stories.js
 __snapshots__
+*.mp4


lib/.npmignore
```diff
@@ -5,4 +5,5 @@ android/build
 screenshots
 *-test.js
 stories.js
-__snapshots__
\ No newline at end of file
+__snapshots__
+*.mp4
```
georgegillams: can we also ignore `screenshots` while we're at it?

tiagoengel: It's already ignored :)

georgegillams: perfect

PR #409, opened 2020-03-11
[BPK-3678] add unit tests for bridges

tiagoengel:
## Summary

This adds two tests for native bridges and a `package.json` script to run them. It is also now part of the release process. 

More tests will be added in subsequent PRs after this is merged.

## Tests

The idea of those tests is is manly to test if the native code is handling the props received from JS correctly. I will also add unit tests to all other classes in another PR.

It's out of the scope of this to add e2e or screenshot testing as those require extra work to configure. 

To test it locally run `npm run android:test`

**NOTE: Android tests will be added to the CI on another PR.**

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


android/settings.gradle
```diff
@@ -6,7 +6,8 @@ rootProject.name = 'native'
 include ':app'
 
 include ':backpack-react-native'
-project(':backpack-react-native').projectDir = new File(rootProject.projectDir, '../node_modules/backpack-react-native/android')
+project(':backpack-react-native').projectDir = new File(rootProject.projectDir, '../lib/android')
+project(':backpack-react-native').setBuildFileName('./build.internal.gradle')
```
tiagoengel: Added this so we can't publish a "clean" build.gradle file with the npm package and not worry about what we add to the internal one.

lib/android/src/androidTest/java/net/skyscanner/backpack/reactnative/rating/BpkRatingViewManagerTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+  @get:Rule
+  val managerRule = ReactViewManagerTestRule(BpkRatingViewManager::class)
+  private val manager: BpkRatingViewManager
+    get() = managerRule.manager
+
+  @Test
+  fun test_title() {
+    val view =  manager.createViewInstance(managerRule.themedContext)
+    manager.updateProperties(view, buildProps("title", JavaOnlyArray.of("low", "med", "high")))
```
tiagoengel: This is inspired by what the RN codebase does, what is basically happening here is we are testing the process of setting the `title` prop from the JS code.

We could have called `setTitle` directly in `BpkRatingViewManager` but that would not guarantee a prop with this name is available in the native side (e.g. if we forget to annotate the method with the prop name)

These are the most important "unit" tests as most of our logic for bridges is around transforming and setting props coming from the JS code. 

The next most important test are e2e, a.k.a visual regression, a.k.a screenshot test. But those will require quite a lot more work to setup.

android/settings.gradle
```diff
@@ -6,7 +6,8 @@ rootProject.name = 'native'
 include ':app'
 
 include ':backpack-react-native'
-project(':backpack-react-native').projectDir = new File(rootProject.projectDir, '../node_modules/backpack-react-native/android')
+project(':backpack-react-native').projectDir = new File(rootProject.projectDir, '../lib/android')
```
tiagoengel: Moved to `lib/android` because using from `node_modules` causes issues with `gradle`. This happens because we install the package from the file system `file:./lib/backpack-react-native` and the way npm installs it, is by creating a link. This causes `gradle` to compile the files twice and it trows a `duplicate` reference error when compiling the tests.

PR #416, opened 2020-03-13
[BPK-3678] Update to mockk and refactor tests

tiagoengel:
# Mockk

This updates the mocking library to [mockk](https://mockk.io/) because of a few issues I had with Mockito, but especially because mockito does not support mocking final classes (classes that cannot be extended), which was forcing us to open the classes just so we could test them.

Another reason for the move is that this lib is much better for Kotlin code.

The tests now will only work with API >= 28 which is already the API used to create the `bpk-avd` emulator so no changes are needed.

# Refactoring

I've found a bug when any of the Android views is mocked, all tests that run after the mock will fail with some weird internal error in the mocking lib (this is true for both mocking libs), so I decided to refactor the code a bit to make the logic of handling state updates and rendering the view less coupled. It's now possible to do this [test](https://github.com/Skyscanner/backpack-react-native/pull/416/files#diff-fa9e16e0dc120f7cec6e29ebc9878d30R226) without having to mock the whole view.

Another benefit is that the logic to handle state update doesn't need to be repeated inside every bridge component anymore.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -196,9 +196,9 @@ You can also run the tests in 'watch mode', which means the process will continu
 <details>
 <summary>Run Android emulators manually</summary>
 
-The setup process detailed in *[Prerequisites](#prerequisites)* created two Android emulators. One with API version 27 and another with 21.
+The setup process detailed in *[Prerequisites](#prerequisites)* created two Android emulators. One with API version 28 and another with 21.
 
-To run these manually, run `npm run android:emulator` or `npm run android:emulator:min` to run API versions 27 and 21 respectively.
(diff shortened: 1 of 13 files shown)


CONTRIBUTING.md
```diff
@@ -196,9 +196,9 @@ You can also run the tests in 'watch mode', which means the process will continu
 <details>
 <summary>Run Android emulators manually</summary>
 
-The setup process detailed in *[Prerequisites](#prerequisites)* created two Android emulators. One with API version 27 and another with 21.
+The setup process detailed in *[Prerequisites](#prerequisites)* created two Android emulators. One with API version 28 and another with 21.
```
tiagoengel: This was outdated, the command to create the avd uses api 28.

PR #420, opened 2020-03-17
Bpk 3758 add dialog bridge tests

georgegillams:

+ [ ] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/ios/nativeTests/RCTBPKDialogManagerTests.m
+++ b/ios/nativeTests/RCTBPKDialogManagerTests.m
@@ -1,0 +1,65 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2020 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <XCTest/XCTest.h>
+
+#import <BackpackReactNative/RCTBPKDialog.h>
+#import <BackpackReactNative/RCTBPKDialogManager.h>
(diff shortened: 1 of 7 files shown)


ios/nativeTests/RCTBPKDialogManagerTests.m
```diff
@@ -0,0 +1,90 @@
...
+@property(nonatomic, strong) RCTBPKDialog *dialogView;
+@property(nonatomic, strong) BPKDialogController *dialogController;
+@end
+
+@implementation RCTBPKDialogManagerTests
+
+- (void)setUp {
+    self.dialogViewManager = [[RCTBPKDialogManager alloc] init];
+    UIView *view = [self.dialogViewManager view];
+    NSAssert([view isKindOfClass:RCTBPKDialog.class], @"Dialog manager should provide a view of type RCTBPKDialog");
```
k0nserv: I would recommend against using `NSAssert` in tests and asserts in `setUp`.  Can we create a dedicated test to assert this fact with `XCTAssert` instead?

ios/nativeTests/RCTBPKDialogManagerTests.m
```diff
@@ -0,0 +1,90 @@
...
+
+    BPKDialogControllerStyle dialogStyleBottomSheet = [RCTConvert BPKDialogControllerStyle:@"bottomSheet"];
+    XCTAssertTrue(dialogStyleBottomSheet == BPKDialogControllerStyleBottomSheet);
+
+    // test that invalid values fall back to a sensible default
+    BPKDialogControllerStyle dialogStyleInvalid = [RCTConvert BPKDialogControllerStyle:@"invalid"];
+    XCTAssertTrue(dialogStyleInvalid == BPKDialogControllerStyleAlert);
+}
+
+- (void)testSettingBasicProperties {
```
k0nserv: Where we have used `performSelector:` here we are relying on a private implementation detail of the controller, this will mean the tests can start failing because of internal refactoring in `BPKDialogController` which isn't great.

ios/nativeTests/RCTBPKDialogManagerTests.m
k0nserv: Can we also check that the delegate gets correctly set?

ios/nativeTests/RCTBPKDialogManagerTests.m
k0nserv: `UIAlertController` exposes these properties and lets you set them after creation so mabye we should just algin with that? It will be  more familiar and let's us test them

PR #421, opened 2020-03-17
[BPK-3762] Flare component

shaundon:
Skeleton component, doesn't do anything useful yet. Because of our new monopackage, if we merge and release this it'll be available to consumers. Should we:

1. Mark it as beta in some way?
1. Don't worry, it won't be documented so people are unlikely to find it.
1. Not merge this PR until the native side is done on at least one platform?

### DIFF
--- a/lib/bpk-component-flare/src/BpkFlare-test.android.js
+++ b/lib/bpk-component-flare/src/BpkFlare-test.android.js
@@ -1,0 +1,38 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2020 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import commonTests from './BpkFlare-test.common';
+
+jest.mock('react-native', () => {
+  const MARSHMALLOW = 23;
+  const reactNative = jest.requireActual('react-native');
+
+  // We rely on BpkTouchableNativeFeedback to handle the ripple effect on older devices
+  // so in this test suite we're assuming API > 23 where the ripple is supported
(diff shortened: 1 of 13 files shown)


lib/bpk-component-flare/src/BpkFlare-test.android.js
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import commonTests from './BpkFlare-test.common';
+
+jest.mock('react-native', () => {
```
tiagoengel: Is this necessary?

lib/bpk-component-flare/src/BpkFlare.js
```diff
@@ -0,0 +1,86 @@
...
+const BpkFlare = (props: Props) => {
+  const { children, insetPaddingMode, style: userStyle, ...rest } = props;
+
+  const style = [styles.base];
+  if (userStyle) {
+    style.push(userStyle);
+  }
+
+  return (
+    <View style={style} {...rest}>
```
tiagoengel: Can we replace the view here with a `NativeFlare` component? The files would be `NativeFlare.ios` and `NativeFlare.android`  and they can just return a view and spread the props for now,

lib/bpk-component-flare/src/BpkFlare-test.android.js
shaundon: That came from the `create-component` script 🤷‍♂

lib/bpk-component-flare/src/BpkFlare.js
shaundon: Sure, will do that

tiagoengel:
About marking as beta, I wouldn't worry, people will not use if is not documented/doesn't do anything

lib/bpk-component-flare/src/BpkFlare-test.android.js
tiagoengel: Ahh, so it's not necessary then :)

shaundon: I'll shut it down!

lib/bpk-component-flare/src/NativeFlare.android.js
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React from 'react';
+import { View } from 'react-native';
+
+const NativeFlare = ({ ...rest }) => <View {...rest} />;
```
tiagoengel: nit: Can just do `NativeFlare = View` and would make flow stop complaining as well

PR #438, opened 2020-03-19
[BPK-3757] Bridge android flare

tiagoengel:
Removed the `insetPadding` support. It's hard to support it on Android because of how this is implemented, plus it's not used in the app. 

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


lib/android/src/main/java/net/skyscanner/backpack/reactnative/flare/BpkFlareViewManager.kt
```diff
@@ -0,0 +1,48 @@
...
+  public override fun createViewInstance(reactContext: ThemedReactContext): RNBpkFlare {
+    return RNBpkFlare(reactContext)
+  }
+
+  override fun getName(): String {
+    return VIEW_NAME
+  }
+
+  override fun addView(parent: RNBpkFlare, child: View, index: Int) {
+    parent.addViewInternal(child, index)
```
tiagoengel: The reason for this is because `BpkFlare` will throw an exception in case more than 1 children is added. Because of how `shadow nodes` work in react even if we wrap items in a single view in JS, it will add more than one here, which breaks the component.

To go around that we add child views into a internal wrapper, and add this wrapper to `BpkFlare`

PR #439, opened 2020-03-20
[BPK-3693] Improve mask behaviour

shaundon:
`tinymask` sometimes formats things strangely when they contain special characters. For example:

```
const maskInstance = new tinymask('9999-9999');

// Works as expected.
maskInstance.mask('01234567') // 0123-4567
maskInstance.mask('0123-456') // 0123-456

// WTF?
maskInstance.mask('012-456') // 012 
```

This manifests in the component in the GIF you see below - when you remove a character from certain places in the string, `tinymask` reformats it and some of the characters disappear.

This PR fixes this by sanitising masked values before they're given to `tinymask`. I've also added a test for it, and made the example in our storybook make use of `useState` to make it shorter.

## Before fix
before-fix

## After fix
after-fix

### DIFF
--- a/lib/bpk-component-text-input/src/BpkTextInput-test.common.js
+++ b/lib/bpk-component-text-input/src/BpkTextInput-test.common.js
@@ -183,6 +183,17 @@ const commonTests = () => {
         expect(testRenderer.toJSON()).toMatchSnapshot();
       });
 
+      it('should mask correctly when given an unusual value to mask', () => {
+        /* The value here simulates entering '1234-5678', then cursoring back
+          and removing the '3'. It should be correctly re-masked to make '1245-678'.
+        */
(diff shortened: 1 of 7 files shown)


lib/bpk-component-text-input/src/BpkTextInput-test.common.js
```diff
@@ -183,6 +183,17 @@ const commonTests = () => {
         expect(testRenderer.toJSON()).toMatchSnapshot();
       });
 
+      it('should mask correctly when given an unusual value to mask', () => {
```
tiagoengel: Can we add a test for the `*` case and also a mask containing `A` or `S`?

tiagoengel:
Too late I guess 😅

PR #449, opened 2020-03-23
[NO-JIRA] Add Github actions workflow

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -1,0 +1,37 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: macos-10.15
+
+    steps:
+    # Checkout the repo
+    - uses: actions/checkout@v2
+
+    - name: brew install
+      run: brew install clang-format
+
+    - name: npm install
+      run: npm ci
+
+    - name: install bundler
+      run: gem install bundler --version 2.0.2
+
+    - name: bundler install
+      run: bundle install
+      working-directory: ./ios
+
+    - name: CocoaPod install
+      run: bundle exec pod install
+      working-directory: ./ios
+
+    - name: CI
+      run: bundle exec rake ci
+      working-directory: ./ios
+
--- a/ios/Rakefile
+++ b/ios/Rakefile
@@ -6,7 +6,7 @@ EXAMPLE_SCHEMA = 'native'
 ANALYZE_FAIL_MESSAGE = '⚠️'
 
 def get_changed_files()
-  `git status --porcelain | grep -Ev "CHANGELOG|UNRELEASED|package-lock.json"`.lines
+  `git status --porcelain | grep -Ev "CHANGELOG|UNRELEASED|Podfile.lock|package-lock.json"`.lines
 end
 
 def check_pristine()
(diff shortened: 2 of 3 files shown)


.github/workflows/ios-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: macos-10.15
+
+    steps:
+    # Checkout the repo
+    - uses: actions/checkout@v2
```
shaundon: What does this mean? Like are there docs I can view? It's all good but I'm just trying to wrap my head around how GH Actions works

georgegillams: https://github.com/actions/checkout#checkout-v2

.travis.yml
```diff
@@ -1,107 +0,0 @@
-language: node_js
-
-env:
-  global:
-    - NPM_CONFIG_LOGLEVEL=warn
-    - CLOUDSDK_CORE_DISABLE_PROMPTS=1
-    - ANDROID_API_LEVEL=29
```
olliecurtis: Don't we still need the Android CI right now? Deleting this shouldn't be until we migrate

.github/workflows/ios-ci.yml
shaundon: Thanks

.travis.yml
georgegillams: Good catch. I deleted this while I was working just to reduce the load on travis. Will reinstate

ios/Rakefile
```diff
@@ -28,9 +28,8 @@ end
 desc "Runs clang-format on all h and m files to ensure formatting is correct"
 task :lint do
   abort red "The following generated files have changed during setup:\n#{get_changed_files}" unless check_pristine
-  `clang-format -i **/*.h **/*.m`
+  `clang-format -i ../**/*.h ../**/*.m`
```
georgegillams: To include files in the `lib` directory too

.github/workflows/ios-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    - name: brew install
+      run: brew install clang-format
+
+    - name: npm install
+      run: |
+        npm ci
+        if [ -d ./lib/node_modules ]; then cp -rf ./lib/node_modules/* ./node_modules/; fi
```
tiagoengel: Why are we doing this? If npm is putting node_modules there I don't think we should be messing with it. It doesn't seem to cause any problems.

georgegillams: It does, because we assume the location of the font files

tiagoengel: Can you try to rebase? @k0nserv deduped our dependencies to fix that.

tiagoengel: Actually the PR is not merged yet https://github.com/Skyscanner/backpack-react-native/pull/451

k0nserv: Yeah I don't think this should be the solution to this issue. The problem is that we expect the font files to be in `node_modules/bpk-svgs` but they ended up nested under `backpack-react-native` at `node_modules/backpack-react-native/node_modules/bpk-svgs`. I ran `npm dedupe bpk-svgs` in my branch which moves them to the root

.github/workflows/ios-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    - name: bundler install
+      run: |
+        gem install bundler --version 1.17.3
```
k0nserv: This is an old version of bundler, can we upgrade to 2.0?

.github/workflows/ios-ci.yml
k0nserv: I just merged #451 so if you rebase on master that should be it

k0nserv:
So I disabled analysis for now since what we are really after here is tests. We need to do more work for analysis because we are getting errors in RN and other libraries atm

PR #454, opened 2020-03-27
[BPK-3680] Use internal backpack

tiagoengel:
## Summary of the changes

- Now if `USE_RELATIVE` is set to true (.env file) the native project will use the `-internal` version of backpack-android which contains relative.
- The internal build of `backpack-react-native` will now also always depend on the `-internal` version of backpack-android, this will eliminate the transitive dependency problems we been having in the app.
- Add `BackpackReactNative.init` function that will configure relative for the RN Android side when the font files are available. This will replace the custom logic we have in the app.
- It is not necessary to download and copy the relative font files for Android to use relative in the local app, as long as `USE_RELATIVE` is true and you are logged in to internal maven, relative will be used.
- While I was at it, updated gradle to latest version :).

I've tested this in the app and it's working as expected.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/android/app/src/main/java/net/skyscanner/backpack/MainApplication.kt
+++ b/android/app/src/main/java/net/skyscanner/backpack/MainApplication.kt
@@ -11,6 +11,7 @@ import com.facebook.react.shell.MainReactPackage
 import com.facebook.soloader.SoLoader
 import com.jakewharton.threetenabp.AndroidThreeTen
 import java.util.Arrays
+import net.skyscanner.backpack.reactnative.BackpackReactNative
 import net.skyscanner.backpack.reactnative.calendar.CalendarPackage
 import net.skyscanner.backpack.reactnative.dialog.DialogPackage
 import net.skyscanner.backpack.reactnative.flare.BpkFlarePackage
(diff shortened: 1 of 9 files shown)


android/app/src/main/java/net/skyscanner/backpack/MainApplication.kt
```diff
@@ -49,6 +50,7 @@ class MainApplication : Application(), ReactApplication {
     override fun onCreate() {
         super.onCreate()
         AndroidThreeTen.init(this)
+        BackpackReactNative.init(this)
```
tiagoengel: This is what we will use in the app as well, it has the added benefit of us using the same mechanism and potentially finding problems before they happen in the app.

lib/android/src/main/java/net/skyscanner/backpack/reactnative/BackpackReactNative.kt
```diff
@@ -0,0 +1,68 @@
...
+import android.util.Log
+import com.facebook.react.views.text.ReactFontManager
+
+/**
+ * [BackpackReactNative] is the entry point of backpack-react-native for Android.
+ */
+class BackpackReactNative {
+  companion object {
+    private const val TAG = "BackpackReactNative"
+    private val RELATIVE_MAPPINGS = mapOf(
```
tiagoengel: Keys are the font names that will be available to RN and values are the font names available in the backpack-android package

PR #464, opened 2020-04-01
[BPK-3684] Override backpack-android version using env

tiagoengel:
This will make it easier to trigger a RN build from the Android repo and check tests pass against the latest version.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -56,12 +56,12 @@ allprojects {
     }
 
     // Force dependency resolution of backpack-android to a specific version
-    if (project.hasProperty("backpack-version")) {
+      if (System.getenv("OVERRIDE_BACKPACK_VERSION")) {
         afterEvaluate { project ->
             project.configurations.all {
                 resolutionStrategy.eachDependency { DependencyResolveDetails details ->
                     if (details.requested.name == "backpack-android") {
-                        details.useVersion project["backpack-version"]
+                        details.useVersion System.getenv("OVERRIDE_BACKPACK_VERSION")
                     }
                 }
             }


android/build.gradle
```diff
@@ -56,12 +56,12 @@ allprojects {
...
 
     // Force dependency resolution of backpack-android to a specific version
-    if (project.hasProperty("backpack-version")) {
+      if (System.getenv("OVERRIDE_BACKPACK_VERSION")) {
         afterEvaluate { project ->
             project.configurations.all {
                 resolutionStrategy.eachDependency { DependencyResolveDetails details ->
                     if (details.requested.name == "backpack-android") {
-                        details.useVersion project["backpack-version"]
+                        details.useVersion System.getenv("OVERRIDE_BACKPACK_VERSION")
```
georgegillams, line 64: Should this still use `project["backpack-version"]` if no env value is supplied?

georgegillams: If not then I guess it's not really overriding, so maybe it should be `BACKPACK_VERSION` instead of `OVERRIDE_BACKPACK_VERSION`??

tiagoengel: No, `project["backpack-version"]` was the previous way we were doing it, it requires providing that to `gradle` (e.g. `./gradlew -Pbackpack-version=<version> <commands>`). 

But it's easier to use an env var to do this on travis, and they serve the same purpose, there's no need to have both.

tiagoengel: it is `override`, this will override the version defined in the project. This is so we can check compiling and testing agains the latest version works.

tiagoengel: This is the relevant Android PR @georgegillams https://github.com/Skyscanner/backpack-android/pull/434/files#diff-2dc095a144b806ada6ae6c51063b3dc9R17

PR #477, opened 2020-04-08
[BPK-3301] Bridge android snackbar

tiagoengel:
Android code was mostly based on https://github.com/cooperka/react-native-snackbar/tree/master/android.

This is not a traditional react component that is rendered inside the component tree, it works instead more like the `Alert` component. 

To render a Snackbar the user will call `BpkSnackbar.show(..)` in response to an event, and `BpkSnackbar.dismiss()` to dismiss it.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


lib/bpk-component-snackbar/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```js
+import React from 'react';
+import { Alert } from 'react-native';
+import Buttom from 'backpack-react-native/bpk-component-buttom';
+import { icons } from 'backpack-react-native/bpk-component-icon';
+import BpkSnackbar, { LENGTH_SHORT } from 'backpack-react-native/bpk-component-snackbar';
+
+export default () => (
+  <Buttom
```
shaundon, line 19: Typo:

```suggestion
import Button from 'backpack-react-native/bpk-component-button';
import { icons } from 'backpack-react-native/bpk-component-icon';
import BpkSnackbar, { LENGTH_SHORT } from 'backpack-react-native/bpk-component-snackbar';

export default () => (
  <Button
```

Also, shouldn't it be `BpkButton`, we still use prefixes right?

lib/bpk-component-snackbar/README.md
```diff
@@ -0,0 +1,56 @@
...
+    onPress={() => {
+      BpkSnackbar.show({
+        text: 'Snackbar text',
+        duration: LENGTH_SHORT,
+        title: 'Title',
+        icon: icons.flight,
+        action: {
+          text: 'Action',
+          onPress: () => {
+            Alert.alert('Action pressed');
```
shaundon, line 30: We should maybe use `BpkAlert` here to set a good example. API is the same as it's just a wrapper around RN's `Alert` component.

lib/bpk-component-snackbar/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+| Property         | PropType                                       | Required | Default Value |
+| ---------------- | ---------------------------------------------- | -------- | ------------- |
+| `duration`       | `number`                                       | true     | `LENGTH_SHORT`|
+| `text`           | `string`                                       | true     | -             |
+| `title`          | `string`                                       | false    | -             |
+| `icon`           | `string`                                       | false    | -             |
+| `action`         | `object`                                       | false    | -             |
+| `action.text*`   | `string`                                       | true     | -             |
+| `action.icon*`   | `string`                                       | true     | -             |
```
shaundon, line 51: What's the deal with the `*`s here?

lib/bpk-component-snackbar/README.md
```diff
@@ -0,0 +1,56 @@
...
+| `action.onPress` | `function`                                     | true     | -             |
+
+**NOTES**
+
+- when `action` is provided `action.text` or `action.icon` should be provided, but not both at the same time.
```
shaundon, line 56: I'm wondering if there's a less surprising way to do the API here.. 

Would there be much downside to making it three separate props, with warnings displayed if people try to use them incorrectly?

I guess it would be like:

```
onActionButtonPressed: func
actionButtonText: string
actionButtonIcon: BpkIcon

Note: When `onActionButtonPressed` is provided you must also pass in either `actionButtonText` or `actionButtonIcon`. You cannot pass both.
``` 

And then in the code throw warnings/errors if they pass in both `actionButtonText` and `actionButtonIcon`.

I realise this might be more difficult/impossible due to the bridging, so only if it's actually feasible.

lib/bpk-component-snackbar/src/NativeSnackbar.android.js
```diff
@@ -0,0 +1,54 @@
...
+
+import { type INativeSnackbar } from './common-types';
+
+const { AndroidBpkSnackbar } = NativeModules;
+
+export const {
+  LENGTH_SHORT,
+  LENGTH_LONG,
+  LENGTH_INDEFINITE,
+} = AndroidBpkSnackbar;
```
shaundon, line 31: Is there any way we can rename these to be more explicit? 

We usually tie in the component name, for example `BUTTON_TYPES` or `BADGE_ACCESSORY_VIEW_ALIGNMENTS`. So it'd be good to make them `SNACKBAR_LENGTHS.short` and so on (note the use of plural too). It'd also be good if they were in one `SNACKBAR_LENGTHS` object too.

Adding the component name would also avoid any potential future namespace clashes if we move to requiring all components from a single Backpack import.

I realise this might be more difficult due to the bridge, I'm just assuming there's a way to map/remap these names before they go over the bridge.

shaundon:
Runs great for me in Storybook, although RTL didn't seem to work for me?

lib/bpk-component-snackbar/README.md
tiagoengel: Just to note there is more information about it bellow. Not sure how to make this more clear

shaundon: Ahh, I was thinking it from a regex/glob pattern point of view 😂

lib/bpk-component-snackbar/README.md
tiagoengel: Honestly I think different props is worse as it gives the impression you can have both an action button and a text button at the same time, which is not true. 

Plus the way types are defined now flow will ensure it is impossible to set `text` and `icon` at the same time, following this proposed api I'm not sure that will be possible.

shaundon: Yeah fair enough. Let's leave it this way then

lib/bpk-component-snackbar/src/NativeSnackbar.android.js
tiagoengel: It's possible, I'm not a big fan of this type of api to be honest, it's unnecessary noise in JS. Plus if it ever clash people can alias the import. But I if this is what we generally to is better to be consistent.

tiagoengel:
You have to refresh the app after you toggle RTL in storybook, I'll have a look, but If is not working it's probably a problem in the native component.

lib/bpk-component-snackbar/src/NativeSnackbar.android.js
shaundon: Yeah I feel like consistency is what we should go for here

shaundon:
> You have to refresh the app after you toggle RTL in storybook, I'll have a look, but If is not working it's probably a problem in the native component.

Thanks, I did refresh it and I tested RTL was working for other components, so yeah might be a bug on the native side - in which case we don't need to block this because it can be fixed on Android

tiagoengel:
> Should it be `hide` instead of `dismiss`? I'm just thinking that as `show` and `hide` are opposites, it would be more 'principle of least surprise' to use this terminology

`dismiss` follows the naming on iOS and Android, it's more `dismissing` a notification, as a snackbar is something show to notify you, and you are then "dismissing" that notification and not just hiding it.

tiagoengel:
Tested RTL using an RTL language and works fine, it's probably a bug with the RTL force option in RN

Screenshot_1586512797

PR #478, opened 2020-04-08
[BPK-3800] Upgrade most Storybook dependencies

shaundon:
[This issue](https://github.com/storybookjs/react-native/issues/24) is preventing us from upgrading `react-native-server` but I was able to upgrade the rest. I'm now watching that issue so can try upgrading once it's resolved.

### DIFF
--- a/storybook/storybook.js
+++ b/storybook/storybook.js
@@ -19,7 +19,7 @@
 /* @flow */
 import React, { type Node } from 'react';
 import addon from '@storybook/addons';
-import { I18nManager, YellowBox, View } from 'react-native';
+import { I18nManager, View } from 'react-native';
 import { getStorybookUI, configure } from '@storybook/react-native';
 import { backgroundColor } from 'bpk-tokens/tokens/base.react.native';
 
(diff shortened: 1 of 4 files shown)


storybook/storybook.js
```diff
@@ -61,12 +61,6 @@ const initDarkModeAddon = channel => {
   channel.on(DM_EVENT, colorScheme => BpkAppearance.set({ colorScheme }));
 };
 
-const hideWarnings = () => {
-  // TODO: this warning is being trigger by an internal react code, we can remove it when it gets fixed
-  // see: https://github.com/facebook/react-native/issues/18868#issuecomment-382671739
-  YellowBox.ignoreWarnings(['Warning: isMounted(...) is deprecated']);
-};
-
```
shaundon, line 69: I checked and this doesn't seem to be an issue anymore, so I removed it.

storybook/storybook.js
```diff
@@ -109,9 +103,12 @@ configure(() => {
 }, module);
 /* eslint-enable global-require */
 
-const StorybookUI = getStorybookUI({ onDeviceUI: false });
+const StorybookUI = getStorybookUI({
+  onDeviceUI: false,
+  asyncStorage: null,
```
shaundon, line 108: See https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#react-native-async-storage.

I tried to get it working but it caused a world of hurt, and for such a small feature I thought it's not worth the hassle. We can revisit this in the future if necessary.

PR #487, opened 2020-04-14
[BPK-3810] Add pointerDirection prop to flare component

shaundon:
Baby's first bridge!

I added some extra margin above the screenshots to make them look better.


lib/ios/BackpackReactNative/Classes/FlareBridge/RCTBPKFlareManager.m
```diff
@@ -28,4 +29,6 @@ - (UIView *)view {
     return [[RCTBPKFlare alloc] initWithFrame:CGRectZero];
 }
 
+RCT_REMAP_VIEW_PROPERTY(pointerDirection, flarePosition, BPKFlarePosition)
```
georgegillams, line 32: I have a feeling we generally prefix native properties that are mapped to RN with `rct_`. Maybe that's not always the case though
```suggestion
RCT_REMAP_VIEW_PROPERTY(pointerDirection, rct_flarePosition, BPKFlarePosition)
```

lib/bpk-component-flare/index.js
```diff
@@ -18,7 +18,8 @@
 
 /* @flow */
 
-import BpkFlare from './src/BpkFlare';
+import BpkFlare, { FLARE_POINTER_DIRECTIONS } from './src/BpkFlare';
 
 export type { Props as BpkFlareProps } from './src/BpkFlare';
 export default BpkFlare;
+export { FLARE_POINTER_DIRECTIONS };
```
georgegillams, line 25: I wonder if `pointerDirection` could be confusing in the future if we add support for other variants. For example 
image

Maybe `pointerPosition` or something is better?

lib/ios/BackpackReactNative/Classes/FlareBridge/RCTBPKFlareManager.m
shaundon: I copied it from Dialog which isn't doing that.

My understanding of the API is:

```
RCT_REMAP_VIEW_PROPERTY(
  name_to_use_in_js,
  name_of_property_in_ios,
  type
)
```

If I'm correct we can't use anything other than `flarePosition` because that's the name of the property in iOS

lib/bpk-component-flare/index.js
shaundon: Gonna CC @tiagoengel here as we came up with the name together

tiagoengel: Pointer `direction` is what we use on Android for vertical direction. I think `position` is more generic and it is also what we use on Android for horizontal position, `middle|start|end`. 

I also think that `direction` is more accurate in this case as we are talking about the direction the pointer is "pointing", not really the position.

If we want to be really explicit about it we should call `pointerVerticalDirection` or something like that?

tiagoengel: To add `position` and `direction` can be used together on Android, e.g. pointing up and at the start, or pointing down and at the middle, etc

shaundon: I would just go with `direction` tbh @georgegillams. If we add more stuff later we'll figure it out.

georgegillams: Sounds good to me 👍

lib/ios/BackpackReactNative/Classes/FlareBridge/RCTBPKFlareManager.m
georgegillams: Ah yeah you're correct. We would use `rct_flarePosition` if we needed a middle value between js and the native code.

PR #491, opened 2020-04-15
[BPK-3842] Fix a bug where changing the 'mask' prop didn't work

shaundon:
Pretty straightforward fix. `tinymask` was only being created once, so rerenders triggered by React when props changed had no effect. This moves it to the render function and uses lodash's memoize (thanks @tiagoengel!) for efficiency.

I tested this locally in Storybook but I didn't add an example or test for it, as this is standard functionality that shouldn't regress in future (feel free to disagree, maybe I'm wrong)

### DIFF
--- a/lib/bpk-component-text-input/src/BpkTextInput.js
+++ b/lib/bpk-component-text-input/src/BpkTextInput.js
@@ -23,6 +23,7 @@ import React, { Component, type Node, type Config } from 'react';
 import { Animated, TextInput, View, ViewPropTypes } from 'react-native';
 import AnimatedValue from 'react-native/Libraries/Animated/src/nodes/AnimatedValue';
 import { animationDurationSm } from 'bpk-tokens/tokens/base.react.native';
+import memoize from 'lodash/memoize';
 import TinyMask from 'tinymask';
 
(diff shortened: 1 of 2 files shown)


lib/bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -124,11 +125,11 @@ type State = {
 
 type EnhancedProps = Props & WithBpkAppearanceInjectedProps;
 
+const createTinyMask = memoize(mask => new TinyMask(mask || ''));
```
MichaelTkachuk, line 128: I believe it might be good to avoid having the same memoize cache for all the instances of BpkTextInput, as cache will be constantly overwritten by different instances of BpkTextInput.

Perhaps createTinyMask should be declared as a BpkTextInput method, this will ensure that separate memoize cache is created per component instance.

tiagoengel: The cache is per "mask" string, so there is no issue in using the same, it's actually beneficial to do so as all mask strings will share the same instance of `TinyMask`.

tiagoengel:
Code looks good, I'll try it locally

PR #520, opened 2020-05-19
BpkText styles inheritance support

MichaelTkachuk:
+ [ ] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-react-native/blob/master/CONTRIBUTING.md)

Remember to include the following changes:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests
+ [x] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)

_If you are curious about how we review, please read through the [code review guidelines](https://github.com/Skyscanner/backpack/blob/master/CODE_REVIEW_GUIDELINES.md)_

PR #521, opened 2020-05-19
[BPK-2527]: Added text link component

olliecurtis:
New BpkTextLink components

default
default_dm

default_dm
default


lib/bpk-component-text-link/src/BpkTextLink.ios.js
```diff
@@ -0,0 +1,74 @@
...
+  const themeAttributes = getThemeAttributes(REQUIRED_THEME_ATTRIBUTES, theme);
+
+  const themeStyle = {
+    text: themeAttributes && {
+      color: themeAttributes.textLinkColor,
+    },
+  };
+
+  return (
+    <TouchableWithoutFeedback
```
olliecurtis, line 54: Using this directly instead of `BpkTouchableOverlay` as it was causing the link not to render when passing children into it.

The Backpack component is just a wrapper around this component anyway so given the basic requirements we need I felt this could work as I spent 4days trying to work out why the Backpack component didn't work but couldn't see

lib/bpk-component-text-link/src/BpkTextLink-test.android.js
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import commonTests from './BpkTextLink-test.common';
+
+jest.mock('react-native', () => {
```
tiagoengel, line 23: This can be removed

lib/bpk-component-text-link/src/common-types.js
```diff
@@ -0,0 +1,44 @@
...
+import { type Node } from 'react';
+import PropTypes from 'prop-types';
+
+import { type Theme } from '../../bpk-theming';
+import { themePropType } from '../../bpk-component-text/src/theming';
+
+export type Props = {
+  children: Node,
+  onPress: () => mixed,
+  accessibilityLabel: ?string,
```
tiagoengel, line 30: Shall we make this required?

lib/bpk-component-text-link/src/BpkTextLink.ios.js
```diff
@@ -0,0 +1,74 @@
...
+      accessibilityLabel={accessibilityLabel}
+      accessibilityRole="link"
+      onPress={onPress}
+    >
+      <BpkText style={[styles.text, themeStyle.text]} {...rest}>
```
tiagoengel, line 59: Wow would this work when it's inside a text with a different `textStyle` or `weight`? I believe the current implementation will not work as expected, it will show the link using the `base` text style even if it's inside a `lg` text, for example. We should probably add a story for it.

I think the work being done here is more or less what we need to make this work https://github.com/Skyscanner/backpack-react-native/pull/520

olliecurtis: I guess the idea here when we don't consider #520 is that consumers would pass the styles like they would for `BpkText` into this component, but I guess in the default case where users only apply the style to the parent `BpkText` it would be different as mentioned 😄

lib/bpk-component-text-link/src/common-types.js
olliecurtis: Sounds good to me 😄

lib/bpk-component-text-link/src/BpkTextLink-test.android.js
olliecurtis: Just the `jest.mock` right? 

Will also make a separate PR after this to remove this from the boilerplate lib if its no longer needed 😄

lib/bpk-component-text-link/src/BpkTextLink.ios.js
tiagoengel: If that is the case we need to extend the text props for this component as well. 

```js

type Props = $Exact<TextProps> & {
  <linkProps>
}
```

And the same for PropTypes

lib/bpk-component-text-link/src/BpkTextLink-test.android.js
tiagoengel: yep, just the mock. It's not necessary to mock the platform anymore, it's based on the file extension. `.android` or `.ios` :)

lib/bpk-component-text-link/src/BpkTextLink.ios.js
olliecurtis: Ok that sounds good to me - will make that change 😄

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,72 @@
...
+import BpkTextLink from './index';
+
+storiesOf('bpk-component-text-link', module)
+  .addDecorator(CenterDecorator)
+  .add('default', () => (
+    <BpkText>
+      Hello World! This{' '}
+      <BpkTextLink
+        accessibilityLabel="Link to press"
+        onPress={() => console.log('Clicked the link')}
```
tiagoengel, line 40: Can we use an `action` here instead of `console.log`? 

```
import action from '../../storybook/addons/actions';
...
onPress={action("Clicked")}
```

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,72 @@
...
+      >
+        link
+      </BpkTextLink>{' '}
+      is clickable!
+    </BpkText>
+  ))
+  .add('with-mixed-styles', () => (
+    <BpkText textStyle="lg" weight={WEIGHT_STYLES.heavy}>
```
tiagoengel, line 48: This is using the wrong combination of style and weight, `lg` is not supported with `heavy` weight. Let's change this to emphasised. I also think we should change this story to use `lg` and `emphasized` for the link as well, I don't see this case being very common, if the link is in the middle of the text it would normally have the same style. I would change this to `With different text style`

lib/bpk-component-text-link/src/BpkTextLink.ios.js
tiagoengel: Seeing how this behave in practice I'm not sure this is the best approach, the react docs mention `Do not use unless you have a very good reason. All elements that respond to press should have a visual feedback when touched.`

Also, `BpkTouchableOverlay` as far as I remember is supposed to be used only on iOS, Android has `BpkTouchableNativeFeedback` for the same purpose.

Check the implementation of button for example, iOS is using BpkTouchableOverlay and Android BpkTouchableNativeFeedback, so I believe we should stick to the same approach.

lib/bpk-component-text-link/stories.js
olliecurtis: Sure thing! Guessing that will solve the error right?

lib/bpk-component-text-link/stories.js
olliecurtis: Sure thing no problem!

lib/bpk-component-text-link/README.md
```diff
@@ -0,0 +1,51 @@
+# bpk-component-text-link
+
+> Backpack React Native text-link component.
```
shaundon, line 3:
```suggestion
> Backpack React Native text link component.
```

lib/bpk-component-text-link/README.md
```diff
@@ -0,0 +1,51 @@
...
+    flex: 1,
+    justifyContent: 'center',
+    padding: spacingBase,
+  }
+});
+
+export default () => (
+  <View style={styles.container}>
+    <BpkText>
+        Check out the website <BpkTextLink accessibilityLabel="link to website" onPress={() => {}}>here</BpkTextLink>
```
shaundon, line 28: My pet peeve is links that say `here`, links should be like road signs that tell you where they lead:

```suggestion
        Check out the <BpkTextLink accessibilityLabel="link to website" onPress={() => {}}>website</BpkTextLink>
```

lib/bpk-component-text-link/README.md
```diff
@@ -0,0 +1,51 @@
...
+      </BpkText>
+  </View>
+);
+```
+
+## Props
+
+### BpkTextLink
+
+This component creates a clickable instance of the `BpkText` component and accepts the same props as [`BpkText`](/components/text/?platform=native), alongside the following:
```
shaundon, line 38: Touchscreens don't click*, this is backed up by the event handler being named `onPress`.

*Many tablets do have mouse support but they're still simulating a press.

```suggestion
This component creates a pressable instance of the `BpkText` component and accepts the same props as [`BpkText`](/components/text/?platform=native), alongside the following:
```

lib/bpk-component-text-link/README.md
```diff
@@ -0,0 +1,51 @@
...
+| accessibilityLabel  | string                    | true     | null          |
+| children            | node                      | true     | -             |
+| onPress             | func                      | true     | -             |
+| theme               | See [Theme Props](#theme-props) below     | false    | null          |
+
+
+
+## Theme Props
+
+* `linkTexColor`
```
shaundon, line 51:
```suggestion
* `linkTextColor`
```

lib/bpk-component-text-link/src/BpkTextLink-test.common.js
```diff
@@ -0,0 +1,76 @@
...
+import BpkTextLink from './BpkTextLink';
+
+const commonTests = () => {
+  describe('BpkTextLink', () => {
+    it('should render correctly', () => {
+      const tree = renderer
+        .create(
+          <BpkTextLink
+            accessibilityLabel="Tappble link"
+            onPress={() => console.log('Clicked the link')}
```
shaundon, line 34:
```suggestion
            onPress={() => console.log('Pressed the link')}
```

lib/bpk-component-text-link/src/BpkTextLink-test.common.js
```diff
@@ -0,0 +1,76 @@
...
+
+import BpkTextLink from './BpkTextLink';
+
+const commonTests = () => {
+  describe('BpkTextLink', () => {
+    it('should render correctly', () => {
+      const tree = renderer
+        .create(
+          <BpkTextLink
+            accessibilityLabel="Tappble link"
```
shaundon, line 33:
```suggestion
            accessibilityLabel="Tappable link"
```

lib/bpk-component-text-link/src/BpkTextLink-test.common.js
```diff
@@ -0,0 +1,76 @@
...
+        )
+        .toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly with arbitrary props', () => {
+      const tree = renderer
+        .create(
+          <BpkTextLink
+            accessibilityLabel="Tappble link"
```
shaundon, line 47:
```suggestion
            accessibilityLabel="Tappable link"
```

lib/bpk-component-text-link/src/BpkTextLink-test.common.js
```diff
@@ -0,0 +1,76 @@
...
+            onPress={() => console.log('Clicked the link')}
```
shaundon, line 48:
```suggestion
            onPress={() => console.log('Pressed the link')}
```

lib/bpk-component-text-link/src/BpkTextLink-test.common.js
```diff
@@ -0,0 +1,76 @@
...
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should support theming', () => {
+      const theme = { linkTextColor: 'red' };
+      const tree = renderer
+        .create(
+          <BpkThemeProvider theme={theme}>
+            <BpkTextLink
+              accessibilityLabel="Tappble link"
```
shaundon, line 64:
```suggestion
              accessibilityLabel="Tappable link"
```

lib/bpk-component-text-link/src/BpkTextLink-test.common.js
```diff
@@ -0,0 +1,76 @@
...
+              onPress={() => console.log('Clicked the link')}
```
shaundon, line 65:
```suggestion
              onPress={() => console.log('Pressed the link')}
```

lib/bpk-component-text-link/src/BpkTextLink.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  if (style) {
+    textStyles.push(style);
+  }
+
+  return (
+    // We are using TouchableWithoutFeedback as there is incompatibility with embedding Non Text elements within text which causes errors
+    // on Android and iOS devices
+    <TouchableWithoutFeedback
```
shaundon, line 68: Does this mean there is no visual feedback when this component is pressed on? 

I think that's an acceptable loss in light of your comment about it not working otherwise, I'm just trying to understand the tradeoff.

When we get to RN 0.63 we might be able to use [`Pressable`](https://reactnativejobs.com/blog/react-native-pressable-platformcolor-dynamiccolor/#pressable) to add visual feedback, which would fix this.

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,75 @@
...
+import BpkTextLink from './index';
+
+storiesOf('bpk-component-text-link', module)
+  .addDecorator(CenterDecorator)
+  .add('default', () => (
+    <BpkText>
+      Hello World! This{' '}
+      <BpkTextLink
+        accessibilityLabel="Link to press"
+        onPress={action('Clicked the link')}
```
shaundon, line 41:
```suggestion
        onPress={action('Pressed the link')}
```

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,75 @@
...
+      >
+        link
+      </BpkTextLink>{' '}
+      is clickable!
```
shaundon, line 45:
```suggestion
      is pressable!
```

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,75 @@
...
+    </BpkText>
+  ))
+  .add('with-different-text-style', () => (
+    <BpkText textStyle="lg" weight={WEIGHT_STYLES.emphasized}>
+      Hello World! This{' '}
+      <BpkTextLink
+        accessibilityLabel="Link to press"
+        onPress={action('Clicked the link')}
```
shaundon, line 53:
```suggestion
        onPress={action('Pressed the link')}
```

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,75 @@
...
+        textStyle="lg"
+        weight={WEIGHT_STYLES.emphasized}
+      >
+        link
+      </BpkTextLink>{' '}
+      is clickable!
```
shaundon, line 59:
```suggestion
      is pressable!
```

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,75 @@
...
+    </BpkText>
+  ))
+  .add('with-theme', () => (
+    <BpkThemeProvider theme={themeAttributes}>
+      <BpkText>
+        Hello World! This{' '}
+        <BpkTextLink
+          accessibilityLabel="Link to press"
+          onPress={action('Clicked the link')}
```
shaundon, line 68:
```suggestion
          onPress={action('Pressed the link')}
```

lib/bpk-component-text-link/stories.js
```diff
@@ -0,0 +1,75 @@
...
+        >
+          link
+        </BpkTextLink>{' '}
+        is clickable!
```
shaundon, line 72:
```suggestion
        is pressable!
```

lib/bpk-component-text-link/src/BpkTextLink.js
olliecurtis: It will only on Android, for iOS visual feedback is displayed even though it says TouchableWithoutFeedback 😆

PR #537, opened 2020-06-11
[BPK-3195] Add `forwardedRef` prop to `withTheme` HOC

shaundon:
Helios asked for this, was trivial to add so I went ahead and did it.

Not documented in the readme because `withTheme` was never documented. I'm not even sure why they're using it instead of `BpkThemeProvider` but ¯\\_(ツ)_/¯ 

I confirmed that this doesn't break anything by loading the storybook and checking the themed stories still work.


lib/bpk-component-text/src/BpkText-test.android.js
```diff
@@ -22,11 +22,12 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 
 import BpkThemeProvider from '../../bpk-theming';
-import { WEIGHT_STYLES } from '..';
 
 import BpkText from './BpkText';
 import commonTests from './BpkText-test.common';
 
+import { WEIGHT_STYLES } from '..';
+
```
shaundon, line 30: The linter has decided we need to change this so whatever

lib/bpk-component-text/src/BpkText-test.android.js
```diff
@@ -22,11 +22,12 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 
 import BpkThemeProvider from '../../bpk-theming';
-import { WEIGHT_STYLES } from '..';
 
 import BpkText from './BpkText';
 import commonTests from './BpkText-test.common';
 
+import { WEIGHT_STYLES } from '..';
```
georgegillams, line 29: Can we have `../` instead? Feels more explicit

shaundon: Sure

shaundon: Actually no, I tried doing it and it reverts back to `..` when I save (and the linter/prettier runs). Sorry :P

shaundon: I think I had an old version of the linter installed because this change was reverted once I did `npm i` re-ran it.

ios/Rakefile
```diff
@@ -1,5 +1,5 @@
 FULL_TESTS = ENV['FULL_TESTS'] != 'false'
-BUILD_SDK = ENV['BUILD_SDK'] || 'iphonesimulator13.4'
+BUILD_SDK = ENV['BUILD_SDK'] || 'iphonesimulator13.5'
```
shaundon, line 2: CI was failing so I had to do this change.

```xcodebuild: error: SDK "iphonesimulator13.4" cannot be located.```

File: UNRELEASED.md [hunk 4]
```diff
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Added:**
+
+- bpk-theming:
+  - Added `forwardedRef` prop to `withTheme` HOC, for passing down custom refs to the wrapped component.
```
tiagoengel, line 8: Shall we make this a breaking change? According to react docs adding forwardRef should be treated as a breaking change https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

shaundon: Yeah you might be right. Better safe than sorry anyway, I'll change it.

lib/bpk-theming/src/create-with-theme.js
```diff
@@ -30,6 +31,14 @@ const createWithTheme = () => {
 
   return Component => {
     class WithTheme extends React.Component {
+      static propTypes = {
+        forwardedRef: PropTypes.func,
```
tiagoengel, line 35: Ref is not always a function, the correct type is 
```js
PropTypes.oneOfType([
    // Either a function
    PropTypes.func, 
    // Or the instance of a DOM native element (see the note about SSR)
    PropTypes.shape({ current: PropTypes.instanceOf(Element) })
])
```

UNRELEASED.md
shaundon: Actually, I think we've got it wrong. 

The part of the document that talks about breaking changes is where it talks about using `React.forwardRef`. The part about adding a prop to HOCs - like we are - can't be breaking, as we're simply adding a new prop, it doesn't change any existing behaviour.

lib/bpk-theming/src/create-with-theme.js
shaundon: I copied this from another component for consistency. Will change them both to this format.

lib/bpk-theming/src/create-with-theme.js
```diff
@@ -30,6 +31,14 @@ const createWithTheme = () => {
...
+      };
+
+      static defaultProps = {
+        forwardedRef: null,
```
tiagoengel, line 39: Maybe we should make this `undefined` by default? E.g. if ref is not provided it won't be set. At least we avoid the snapshot pollution with undefined.

shaundon: Good idea, will change.

lib/bpk-theming/src/create-with-theme.js
shaundon: Actually wouldn't this only apply to web? On RN, there is no built in `Element` right? And if it's an instance of a component, components are functions.

UNRELEASED.md
tiagoengel: Hmm, that's true, but I think we are doing things wrong then. We shouldn't be exposing the `forwardedRef` prop, but instead returning `React.forwardRef` at the end. 

```
return React.forwardRef((props, ref) => {
    return <Component {...props} forwardedRef={ref} />;
  });
```
So for people using `withTheme` the returned element will still accept `ref` which is what is expected

lib/bpk-theming/src/create-with-theme.js
shaundon: Done - and reverted snapshot changes as a result 👍

lib/bpk-theming/src/create-with-theme.js
tiagoengel: I think element in this case is `React.Element` which is the same for both. The second version is what `useRef` or `React.createRef` returns.

UNRELEASED.md
shaundon: Ah yeah that makes sense. Will try it

shaundon: I tried it and it broke almost all of our tests:

```
    Invariant Violation: Objects are not valid as a React child (found: object with keys {$$typeof, render}). If you meant to render a collection of children, use an array instead.
        in withTheme(BpkHorizontalNavSelectedIndicator)
        in BpkAppearanceProvider
        in Unknown
```

I feel like we should stick with the original behaviour, given that's what that page specifically recommends for HOCs?

ios/Rakefile
olliecurtis: This is that thing that we discussed previously in the squad where they make a change and sims are no longer available on GH Actions so this is all good by me!

UNRELEASED.md
tiagoengel: How do you mean? this is the example in react docs 
```js
function logProps(Component) {
  class LogProps extends React.Component {
    componentDidUpdate(prevProps) {
      console.log('old props:', prevProps);
      console.log('new props:', this.props);
    }

    render() {
      const {forwardedRef, ...rest} = this.props;

      // Assign the custom prop "forwardedRef" as a ref
      return <Component ref={forwardedRef} {...rest} />;
    }
  }

  // Note the second param "ref" provided by React.forwardRef.
  // We can pass it along to LogProps as a regular prop, e.g. "forwardedRef"
  // And it can then be attached to the Component.
  return React.forwardRef((props, ref) => {
    return <LogProps {...props} forwardedRef={ref} />;
  });
}
```

Note the return of the HOC function is `forwardRef` so consumers don't need to know for this specific component ref is passed using `forwardedRef` which is not a common practice.

shaundon:
Travis isn't reporting back so I need to admin force this merge, but you can see it has passed: https://travis-ci.org/github/Skyscanner/backpack-react-native/builds/697255774

PR #540, opened 2020-06-22
[BPK-4044] More context for `BpkThemeAttributes` deprecation with migration guide

shaundon:
I added a migration guide too to make it easier for people.


lib/bpk-theming/README.md
```diff
@@ -42,3 +42,64 @@ export default class App extends Component {
...
+|-|-|
+| `colorGray50` | `colorSkyGrayTint07` |
+| `colorGray100` | `colorSkyGrayTint06` |
+| `colorGray200` | `colorSkyGrayTint05` |
+| `colorGray300` | `colorSkyGrayTint04` |
+| `colorGray400` | `colorSkyGrayTint03` |
+| `colorGray500` | `colorSkyGrayTint02` |
+| `colorGray700` | `colorSkyGrayTint01` |
+| `colorGray900` | `colorSkyGray` |
+| `primaryColor` | `colorSkyBlue` |
```
olliecurtis, line 104: We might have a semantic color here for support with DM?

shaundon: We do but then that might complicate things as semantic colours require you to be using a semantic wrapper. So for simplicity the code sample preserves the functionality exactly as before.

shaundon: I'll add a note though to say semantic colours exist and use those where possible.

PR #560, opened 2020-07-13
[BBPK-4067] Star rating updates

shaundon:
_Relies on #559 being merged first._

This PR updates the star rating component to use the new icon set. See updated screenshots in the diff for how it looks. I tested it in RTL, all good.


shaundon:
> The Android screenshots look to use a different emulator 

It seems to change for me randomly, I've just given up on the idea of them being totally consistent now. As long as people can see the component it's all good.

PR #582, opened 2020-08-07
[BPK-2792] Replace UNSAFE_componentWillReceiveProps

shaundon:
**To do**

**Done**
- horizontal nav
- banner alert

**Out of scope**
- bpk-theming uses these too, but we cannot remove them until we upgrade the theming package. There's a ticket to do this already, so we should address this as part of that ticket (soon!).

### DIFF
--- a/lib/bpk-component-banner-alert/src/AnimateAndFade.js
+++ b/lib/bpk-component-banner-alert/src/AnimateAndFade.js
@@ -82,7 +82,7 @@ class AnimateAndFade extends Component<Props> {
 
     const isCollapsed = this.props.animateOnEnter || !this.props.show;
 
-    this.shouldRenderHeight = true;
+    this.shouldRenderHeight = this.props.animateOnLeave;
     this.innerViewRef = null;
 
     this.height = isCollapsed ? new Animated.Value(COLLAPSED_HEIGHT) : null;
@@ -101,16 +101,14 @@ class AnimateAndFade extends Component<Props> {
     this.measure(captureHeightAndAnimate);
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps: Props) {
+  componentDidUpdate(prevProps: Props) {
+    this.measure(this.animate);
+
     this.shouldRenderHeight = true;
 
-    if (nextProps.show === this.props.show && this.props.show) {
+    if (this.props.show === prevProps.show && prevProps.show) {
       this.shouldRenderHeight = false;
     }
-  }
-
-  componentDidUpdate() {
-    this.measure(this.animate);
   }
 
   measure = (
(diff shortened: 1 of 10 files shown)


lib/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -82,7 +82,7 @@ class AnimateAndFade extends Component<Props> {
 
     const isCollapsed = this.props.animateOnEnter || !this.props.show;
 
-    this.shouldRenderHeight = true;
+    this.shouldRenderHeight = this.props.animateOnLeave;
```
shaundon, line 85: This line is pretty much the entire thing. I don't really understand this component, so basically these are the steps I followed to come to this conclusion:

---

I moved the stuff from `UNSAFE_componentWillReceiveProps` into `componentDidUpdate`. This worked fine for dismissable alerts but not expandable ones. 

So then I experimented with some of the props we were checking in `componentDidUpdate`, and after a bit of trial and error I discovered that changing the initial value of `shouldRenderHeight` to `false` made expandable alerts work correctly! 

But, **oh no**, now dismissable alerts weren't being removed correctly (the other alerts weren't sliding back up to cover the empty space it leaves).

So, my workaround was to make the initial value dependent on the `animateOnLeave` prop, as dismissable alerts set this to `true`.

This seems to work fine now. I think. This is a really complex component 😅

lib/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -101,18 +101,16 @@ class AnimateAndFade extends Component<Props> {
...
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps: Props) {
+  componentDidUpdate(prevProps: Props) {
+    this.measure(this.animate);
+
     this.shouldRenderHeight = true;
 
-    if (nextProps.show === this.props.show && this.props.show) {
+    if (this.props.show === prevProps.show && prevProps.show) {
```
georgegillams, line 109: Isn't this doing the same as before?

shaundon: Yeah.. isn't that the point? I'm just trying to preserve the original behaviour as closely as possible. Maybe I've misunderstood your question 😅

georgegillams: Also why isn't it just `this.props.show && prevProps.show`?

shaundon: I honestly don't know, I literally just preserved what was already there and just changed it to use `prevProps`. 

Wasn't it you who wrote this component initially? :P

georgegillams: > Wasn't it you who wrote this component initially?
I can neither confirm nor deny that.

Here's the boolean logic:
IMG_5713

georgegillams:
be safe

PR #600, opened 2020-08-25
[NO-JIRA] Revert storybook breaking change

georgegillams:
Turns out the breaking changes were significant. Reverting now to re-enable development. Can look at upgrading again after


olliecurtis (approved):
LGTM - lets create an eng-health task to do the upgrade if the changes are significant

PR #601, opened 2020-08-27
[BPK-4165] Allow banner-alert to accept node or string for message

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

Turns out the iOS font “Didot” has the same emoji problem in RN

image

### DIFF
--- a/lib/bpk-component-banner-alert/stories.js
+++ b/lib/bpk-component-banner-alert/stories.js
@@ -402,6 +402,18 @@ storiesOf('bpk-component-banner-alert', module)
         <BpkBannerAlert
           style={styles.bannerAlert}
           type={ALERT_TYPES.neutral}
+          message={
+            <>
+              Your booking was successful. Your booking reference is{' '}
+              <BpkText textStyle="sm" weight="emphasized">
+                fr33sh4v0c4d0
+              </BpkText>
+            </>
+          }
+        />
+        <BpkBannerAlert
+          style={styles.bannerAlert}
+          type={ALERT_TYPES.neutral}
           message={message}
         />
         <BpkBannerAlert
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,6 +1,8 @@
 # Unreleased
 
 > Place your changes below this line.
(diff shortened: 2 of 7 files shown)


shaundon (changes requested):
The code changes look good but I think the documentation + testing side could be a bit more comprehensive:

* Can we have a snapshot test for using a Node as the message prop?
* The story example is quite specific to the emoji case, which is going to end up being covered by `BpkEmoji` when that's created. How about making the story more generic, something like a `BpkText` with a bit of bold text embedded within? Or having that in addition to the current story.

shaundon (changes requested):
Prop types table in readme needs updating too

lib/bpk-component-banner-alert/stories.js
```diff
@@ -399,6 +399,18 @@ storiesOf('bpk-component-banner-alert', module)
...
     return (
       <View>
+        <BpkBannerAlert
+          style={styles.bannerAlert}
+          type={ALERT_TYPES.neutral}
+          message={
+            <>
+              Your booking was successful. Your booking reference is{' '}
+              <BpkText textStyle="sm" weight="emphasized">
+                fr33sh4v0c4d0
```
shaundon, line 409: omg you used my example absolute madlad

shaundon (changes requested):
still needs a Jest snapshot but otherwise good

lib/bpk-component-banner-alert/stories.js
georgegillams: I loved it

georgegillams: Simulator Screen Shot - iPhone 11 Pro Max - 2020-08-27 at 09 34 54

shaundon: get down to del taco

samstarling: You love to see it

shaundon (approved):

PR #602, opened 2020-08-28
[BPK-4164] Add bpk-emoji

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


lib/bpk-component-emoji/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+> Backpack React Native emoji component.
+
+## Installation
+
+Check the main [Readme](https://github.com/skyscanner/backpack-react-native#usage) for a complete installation guide.
+
+## Usage
+
+Ensure that all emoji are wrapped in `BpkEmoji` so that they do not affect the line-height of the surrounding text.
```
shaundon, line 11:
```suggestion
Ensure that all emoji are wrapped in `BpkEmoji` so that they do not affect the line height of the surrounding text.
```

lib/bpk-component-emoji/src/BpkEmoji.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+/* @flow */
+
+import React from 'react';
+import { Text, ViewPropTypes, StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+  emoji: {
+    fontFamily: 'System',
```
shaundon, line 26: You might need to make this `sans-serif` if `Platform.OS === 'android'` to avoid the Danger error about iOS tokens in snapshots. `System` isn't a font on Android, so the reason it's probably working right now is because Android is quietly falling back to `sans-serif` (I assume)

lib/bpk-component-emoji/stories.js
```diff
@@ -0,0 +1,37 @@
...
+
+import CenterDecorator from '../../storybook/CenterDecorator';
+import BpkText from '../bpk-component-text';
+
+import BpkEmoji from './index';
+
+storiesOf('bpk-component-emoji', module)
+  .addDecorator(CenterDecorator)
+  .add('docs:default', () => (
+    <BpkText style={{ fontFamily: Platform.OS === 'ios' ? 'Didot' : 'System' }}>
```
olliecurtis, line 34: Should we try and be using Relative here for the font family so that we can see how it will look with our font as that is what the consumers would see when using it?

olliecurtis (changes requested):
LGTM - One small question, but just going to test locally

shaundon:
Couple of questions / suggestions:

* I assumed this was going to be an export from the `bpk-component-text` package — emojis are text after all. Does it need to be a separate thing?

* Have you thought about how we'll document this on the website? Tbh that could be a good argument for including it in `bpk-component-text`, because we could just have an extra set of screenshots on the BpkText page, saying something like "due to the way Skyscanner's font works, `BpkEmoji` is available for..", and then we can just extend the readme for `BpkText` to include an emoji example. It'll be more straightforward than having to have a separate docs page, or combine two readmes onto one docs page as we have for some other components.

* Can we document the rationale for this somewhere a bit more? I'm concerned that a future Backpacker will not know what this is used for. I'd recommend we change the text of the storybook example to say something like "this should be aligned correctly with the emoji", and then also put a comment somewhere — readme, or in the code, to explain why this component exists.

lib/bpk-component-emoji/stories.js
georgegillams: Yeah but currently our example app doesn't support it. When it does, our example can just be `<BpkText> Yay ... <BpkEmoji>🎉</BpkEmoji></BpkText>`

olliecurtis: I thought our example app does support Relative already?

https://github.com/Skyscanner/backpack-react-native/blob/master/CONTRIBUTING.md#using-relative-font

lib/bpk-component-text/stories.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/accessible-emoji */
```
georgegillams, line 1: This rule is deprecated, so just gonna ignore for now.

shaundon (approved):

PR #607, opened 2020-09-01
[BPK-4152] Add BpkPriceMarker

shaundon:
ezgif-4-9efbb000bbd3


olliecurtis (commented):
LGTM - after chatting earlier I have seen it running on iOS so won't get it running locally.

One thing with the marker going larger when `focused` can you check the designs to see if the text size increases too? I know for the web designs the font goes the next size up on focus so might be worth checking its the same here

olliecurtis (commented):
Another quick comment - I know this might be tricky from the Android PoV but can we update/create the screenshots for this so that when we add it to the docs site we can add screenshots for examples?

shaundon:
@olliecurtis well caught. I've added iOS screenshots, I'm looking into getting it running on Android now.

olliecurtis (approved):
LGTM! Once we get the Android part working we are golden 😄

PR #614, opened 2020-09-09
[BPK-4178]: New map markers

olliecurtis:
Added new `BpkMapMarker`
Sep-09-2020 14-21-03
Sep-09-2020 14-23-22


shaundon (commented):
Still reviewing but something just occurred to me so will post it now:

What if we named this `BpkPOIMarker` or `BpkIconMarker` or something like that? I'm thinking that it'll pair nicely with `BpkPriceMarker` then, because they both have quite prescriptive names to describe them. `BpkMapMarker` is quite generic. I realise you copied it from the web, but in hindsight maybe the web one was named wrongly and should also be changed, seeing as I'm already about to make a breaking change on it.

Another benefit is that we could rename `MARKER_STATUSES` to `POI_MARKER_STATUSES` which makes it much clearer about which component it should be used with.

lib/bpk-component-map/src/BpkMapMarkerStyles.js
```diff
@@ -0,0 +1,127 @@
...
+  spacingLg,
+} from 'bpk-tokens/tokens/base.react.native';
+
+import { shadows } from '../../bpk-styles';
+
+const sharedStyles = {
+  wrapper: {
+    ...shadows.base(),
+    paddingHorizontal: spacingLg / 4, // eslint-disable-line backpack/use-tokens
+    paddingVertical: spacingLg / 4, // eslint-disable-line backpack/use-tokens
```
shaundon, line 42: Fun fact, you can trick the linter into allowing this with `spacingLg * 0.25`. I think this code is fine, I just noticed it recently 😛

lib/bpk-component-map/src/BpkMapMarkerStyles.js
```diff
@@ -0,0 +1,127 @@
...
+  },
+  pointer: {
+    backgroundColor: colorMonteverde,
+    borderRadius: borderRadiusSm,
+    height: spacingBase,
+    transform: [{ rotate: '45deg' }],
+    width: spacingBase,
+    left: '38%',
+    position: 'absolute',
+    bottom: -3,
```
shaundon, line 56: `38%` and `-3` seem very specific and not tokens. We should at least include a comment to explain why they're set to that, or declare some descriptive constants or something.

Edit: Same for below where `17` is used.

Edit 2: I scrolled down and I see a lot of things like this actually throughout the styles, so same comment for all of those.

lib/bpk-component-map/stories.js
```diff
@@ -115,6 +121,97 @@ const StatefulPriceMarkersExample = () => {
...
+      icon: icons.landmark,
+    },
+    {
+      id: '5',
+      name: 'Kolkata Springs Hotel',
+      latitude: 35.635,
+      longitude: 139.72,
+      price: 'Sold out',
+      disabled: true,
+      icon: icons.landmark,
```
shaundon, line 171: This is totally not a big deal so I don't mind at all if you push back, but it might be more fun if we used some different icons here to give some variety and renamed them from hotels, e.g. "FC Nara Stadium", "Panjin Aquarium", you get the idea

shaundon (changes requested):

lib/bpk-component-map/stories.js
olliecurtis: Sure thing!

lib/bpk-component-map/src/BpkMapMarkerStyles.js
olliecurtis: Ah ok, I think we should maybe look at this and how the linter does it for web as I think for this kind of thing on web we allow

lib/bpk-component-map/src/BpkMapMarkerStyles.js
olliecurtis: Sure thing! I tried to make these use tokens or combinations of them but nothing fitted which is annoying and different size things caused it to look either way off design or screws up the alignment/visual glitches on the pointer/wrapper

shaundon: My concern is that with so many 'magic numbers' involved, it'll cause maintenance issues in the future without context as to why these specific values are being used, we won't know how to change the component without breaking it.

olliecurtis: Oh yeah I am not saying not adding comments I fully agree with that, I was just commenting here for any reviewers right now why 'magic numbers' were used - fully adding comments to the code

shaundon: Ah gotcha!

lib/bpk-component-map/stories.js
olliecurtis: Made some new naming changes and icons to match!

lib/bpk-component-map/src/BpkMapMarkerStyles.js
olliecurtis: Added some comments and also realised I could make some value changes to use tokens too which helps 😄

shaundon (commented):
I think it looks good now, just gonna try it locally

shaundon (commented):
Oh you forgot to update unreleased

shaundon (approved):
I wanna visit The Panjin Lounge it looks fun

PR #615, opened 2020-09-11
[No Jira] Add edge case story for badge

shaundon:
Question came up from a consumer, so I added a storybook example to save time in future. Not intended to be put on the docs site, just for our own reference.

Screen Shot 2020-09-11 at 11 57 39


georgegillams (approved):
Should we maybe throw in an example with some Cyrillic text too as that one seems to come up a fair bit?

shaundon:
> Should we maybe throw in an example with some Cyrillic text too as that one seems to come up a fair bit?

The middle one **is** Cyrillic 😉

PR #631, opened 2020-10-02
Revert "Bump flow-bin from 0.107.0 to 0.134.0 (#623)"

olliecurtis:
This reverts commit abaa881a82e08619cc48f0b492641f63a0d2c74b.

We found that due to our Flow version being much higher than react native it was actually causing us to change types which were actually incorrect and meant that we would be not following correct types.

As part of this investigation and decision we have added a decision log that for all future flow version upgrades we should be aligning our version to the same version as React Native to ensure that we are correctly using their types.

### DIFF
--- a/decisions/flow_versions.md
+++ b/decisions/flow_versions.md
@@ -1,0 +1,11 @@
+# Flow versions
+
+## Decision
+The version of `flow-bin` that we use in Backpack React Native should align to the same version that React Native is using.
+
+
+## Thinking
+The reason for this is newer versions of Flow use different concepts and features that when we make changes within Backpack could actually mean we misuse and violate types that React Native expects.
+
+## Anything else
+**Note:** Any pull requests that Dependabot or Snyk raise to upgrade flow should only be done if the version being updated aligns to the version in React Native. If the version bump is higher the PR should be closed.
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -9,3 +9,7 @@ update_configs:
           # 5.3.x of this dependency stops the addon panel from appearing.
           # See https://github.com/storybookjs/react-native/issues/24
           dependency_name: "@storybook/react-native-server"
+      - match:
+          # Ignoring as per our decision to ignore flow updates
+          # See https://github.com/skyscanner/backpack-react-native/decisions/flow_versions.md
+          dependency_name: "flow-bin"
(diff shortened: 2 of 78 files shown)


decisions/flow_versions.md
```diff
@@ -0,0 +1,11 @@
+# Flow versions
+
+## Decision
+The version of `flow-bin` that we use in Backpack React Native should align to the same version that React Native is using.
```
shaundon, line 4:
```suggestion
The version of `flow-bin` that we use in Backpack React Native should align to the [same version that React Native is using](https://github.com/facebook/react-native/blob/master/.flowconfig).
```

Adding a link just to make it easier for future backpackers to find it

decisions/flow_versions.md
```diff
@@ -0,0 +1,11 @@
...
+
+
+## Thinking
+The reason for this is newer versions of Flow use different concepts and features that when we make changes within Backpack could actually mean we misuse and violate types that React Native expects.
+
+## Anything else
+**Note:** Any pull requests that Dependabot or Snyk raise to upgrade flow should only be done if the version being updated aligns to the version in React Native. If the version bump is higher the PR should be closed.
```
shaundon, line 11: Maybe we should exclude Flow from Dependabot, what do you reckon? That shouldn't block this PR, I'm just wondering

olliecurtis: I did think about this as an option and it made me wonder that if we do that are we likely to forget to ever update flow and then realise we are versions behind?

In my head I thought it might work as a good reminder when the PRs come that we may need to upgrade but I guess it would also create noise by just constantly making PRs we would close.

Both keeping and excluding it are both valid options just not sure what would be best from our maintenance PoV 🤔

decisions/flow_versions.md
olliecurtis: Like this!

decisions/flow_versions.md
shaundon: I guess we could tie it to RN upgrades, so when we upgrade RN we also upgrade Flow? And then it should be trivial to have Danger remind us to do Flow at the same time as RN upgrades

olliecurtis: Thats a good idea actually as newer versions of RN should hopefully be upgrading to the latest or a newer version of flow too. Will add a Dependabot exclude to this PR to keep that from being updated!

shaundon: Noice! We can do the Danger thing separately

k0nserv (approved):

PR #641, opened 2020-10-16
[BPK-4242] Add trailingAccessoryView prop to BpkChip

shaundon:
* Added `trailingAccessoryView` prop
* Removed the internal `dismissible` prop and instead made the dismissible chip pass in the close icon as a trailing accessory view, to make things neater.
* Because of this, the extra layer of indirection we had with `BpkChipWrapper` became redundant, so I removed it (or rather moved its contents to `BpkChip.js`). The diff makes it look like I've changed loads in `BpkChip`, but I actually just deleted that file and then renamed `BpkChipWrapper` to `BpkChip`.


lib/bpk-component-chip/src/BpkChip.js
```diff
@@ -18,37 +18,257 @@
...
+    adjustedTrailingAccessoryView = React.cloneElement(trailingAccessoryView, {
+      style: trailingAccessoryViewItemStyle,
+    });
+  }
 
-  return <BpkChipWrapper {...rest} dismissible={false} />;
+  return (
+    <BpkChipInner
+      accessibilityLabel={accessibilityLabel}
+      disabled={disabled}
```
georgegillams, line 247: This doesn't do anything.

BpkChipInner passes the `disabled` prop directly to `BpkTouchableOverlay` which passes it to `TouchableWithoutFeedback`. Instead, it should be applied like this:
```
<TouchableWithoutFeedback
...
accessibilityState={{
  diabled: true,
}}
```

Maybe for ease `disabled` should be a prop of `BpkTouchableOverlay`

georgegillams (changes requested):

lib/bpk-component-chip/src/BpkChip.js
shaundon: Ah good spot, will change. Is the PR otherwise ok? As this isn't really part of this PR, it's only showing up in the diff because of the file rename

georgegillams (approved):

PR #645, opened 2020-10-22
[BPK-4247] Panel card 12pt corner radius

georgegillams:
This adopts the latest bpk-tokens. To keep the existing 4pt radiuses the same I've replaced all usage of `borderRadiusSm` with `borderRadiusXs`.
For cards and panels, I've moved from `borderRadiusSm` to `borderRadiusXl`.

I don't seem to be able to reliably get the example app to actually use `v35.0.1` because of a conflicting version in `eslint-plugin-backpack`.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.md
```diff
@@ -2,6 +2,8 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+ - Cards and Panels now have a corner radius of `bpk-spacing-md` by default.
```
shaundon, line 6:
```suggestion
 - The default corner radius of cards and panels has been increased from `bpkSpacingXs` to `bpkSpacingMd` (4 to 12).
```

I think this is a bit more informative as it tells them it's increased, and from what to what.

shaundon (changes requested):
All kinds of components have changed in this PR — not just cards and panels, but banner alert, button, carousel indicator, image, map too. Why is this?

georgegillams:
> All kinds of components have changed in this PR — not just cards and panels, but banner alert, button, carousel indicator, image, map too. Why is this?

Because we adopted the breaking changes in bpk-svgs, so to keep the same value we had to replace `borderRadiusSm` with `borderRadiusXs`.

shaundon:
> Because we adopted the breaking changes in bpk-svgs, so to keep the same value we had to replace borderRadiusSm with borderRadiusXs

That makes sense, but because currently we've only changed those tokens for Android, wouldn't that mean the iOS ones are now wrong?

georgegillams:
> > Because we adopted the breaking changes in bpk-svgs, so to keep the same value we had to replace borderRadiusSm with borderRadiusXs
> 
> That makes sense, but because currently we've only changed those tokens for Android, wouldn't that mean the iOS ones are now wrong?

AFAIK we've changed both?
https://github.com/Skyscanner/backpack/blob/master/CHANGELOG.md#2020-10-20

georgegillams:
This is actually breaking as it will force us to adopt latest `bpk-tokens` changes.

UNRELEASED.md
```diff
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Breaking:**
+ - `bpk-tokens@35.0.1` is now required. To migrate to this version, `spacingSm` must be replaced with `spacingXs`.
```
shaundon, line 6: I'm a bit confused by this comment. Is this what we were discussing on Zoom just now?

shaundon:
CI failure for missing fonts seems a bit odd too :-/

UNRELEASED.md
georgegillams: Yeah - although this is only required for Android (as spacingSm === spacingXs in iOS for now)

georgegillams:
> CI failure for missing fonts seems a bit odd too :-/

That's because it's installing the lib deps in `./lib/node_modules` instead of `./node_modules`

UNRELEASED.md
shaundon: Ah ok, I think this could probably be more informative as it's a bit confusing. Our changelog messages should be written for humans — imagine somebody came to you and asked "what's this change?", that's what we should put.

```suggestion
 - Upgraded `bpk-tokens` to `35.0.1`. On Android, this changes `borderRadiusSm` from `4` to `8`. To keep using the old border radius (`4`), change your uses of `borderRadiusSm` to `borderRadiusXs`.
```

UNRELEASED.md
```diff
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Breaking:**
+ - Upgraded `bpk-tokens` to `35.0.1`. On Android, this changes `borderRadiusSm` from `4` to `8`. To keep using the old border radius (`4`), change your uses of `borderRadiusSm` to `borderRadiusXs`.
+
+**Fixed:**
+ - Cards and Panels now have a corner radius of `bpk-spacing-md` by default.
```
shaundon, line 9:
```suggestion
 - The default corner radius of cards and panels has been increased from `bpkBorderRadiusXs` to `bpkBorderRadiusXs ` (4 to 12).
```

I think this is a bit more informative as it tells them it's increased, and from what to what.

PR #654, opened 2020-10-26
[BPK-4246]: Button radius updates

olliecurtis:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


package.json
```diff
@@ -83,12 +83,13 @@
     "babel-loader": "^8.0.6",
     "babel-plugin-jest-hoist": "^26.2.0",
     "backpack-react-native": "file:./lib",
+    "bpk-svgs": "^12.1.16",
```
olliecurtis, line 86: Added this in the root as a devDependency as it is not getting installed through `lib/` which then causes iOS CI and local to fail as the module is not installed which is required to copy the svg font files over from the npm package to the iOS build folder

georgegillams: Is this hack needed in CI, or just locally?

olliecurtis: Both

georgegillams (approved):

package.json
olliecurtis: I guess this is something to explore as part of the ticket we have to investigate running locally

PR #657, opened 2020-10-28
[BPK-4246]: BpkTouchableOverlay update

olliecurtis:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.md
```diff
@@ -2,6 +2,9 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+  - `BpkTouchableOverlay` - fixed radius value to align with new `sm` radius from `xs`
```
shaundon, line 6: This isn't very readable to a consumer

```suggestion
  - `BpkTouchableOverlay` - updated the `sm` value for `borderRadius` to be `8` (previously `4`).
```

shaundon (changes requested):
Cards and panels now use `12` for their border radius, so making this `8` will probably look wrong. We should add an `md` option to this component's `borderRadius` prop and then use it for cards and panels.

shaundon:
It's both — the work you did on this ticket is correct, but I noticed that cards and panels are now using the wrong radius due to them becoming 12 in a previous update. That should have been done as part of the ticket to update them to 12 originally, but it was missed. So it's not that you didn't do the ticket correctly, it's just that I noticed something was broken while reviewing it

lib/bpk-component-card/src/BpkCard.ios.js
```diff
@@ -114,7 +114,9 @@ const BpkCard = (props: Props) => {
     <BpkTouchableOverlay
       accessibilityRole="button"
       style={style}
-      borderRadius={cornerStyle}
+      // We set sm to 'md' here so that the overlay corner styles selected match the card borderRadius
+      // Rather than introducing a new style which is not in the design system we are modifying the default to align correctly.
```
shaundon, line 118: Nice comment

shaundon (approved):
Didn't run locally but I'm confident from the diff that it's all good

PR #661, opened 2020-10-29
[BPK-4251]: Fix local installs by moving bpk to peer

olliecurtis:
Moving bpk dependencies to peer instead of as a direct dependency of Backpack React Native. 

The reasoning for this is so that there aren't any version mis-matches between projects requiring Backpack React Native and using things such as `bpk-tokens` and `bpk-svgs` directly for non Backpack components.

By moving to this model it means that 'dependency hell' is avoided whereby different versions are required by different libraries.


lib/package.json
```diff
@@ -13,15 +13,15 @@
...
-    "bpk-svgs": "^12.1.14",
-    "bpk-tokens": "^36.0.0",
     "lodash": "^4.17.15",
     "recyclerlistview": "^2.0.12",
     "theming": "^1.3.0",
     "tinymask": "^1.0.2"
   },
   "peerDependencies": {
     "@skyscanner/react-native-transitiongroup": "^1.1.3",
+    "bpk-svgs": "^12.1.14",
```
georgegillams, line 23: Don't these need to be installed as devDeps too?

olliecurtis: They are in the root `package.json`, which are also included in this PR 😄

georgegillams: Ignore me - they are!

georgegillams (approved):

georgegillams (approved):

PR #715, opened 2021-01-29
[BPK-4357]: React Native 0.64 

olliecurtis:
This PR upgrades react native to 0.64.1.

Major Changes:

- An important point to this PR everything is upgraded except for Flow in this PR which will be done in a follow up PR and ticket so some `$FlowFixMe` comments have been added and a big block in `untyped` in the `.flowconfig` just to avoid the massive incompatiblity issues both with `react-native` and also with our Skyscanner app.
- Updated CI for effciency
  - Android CI - updated to align more to our Skyscanner/backpack-android and also due to failing gCloud setup on CI which we no longer need anymore since we don't use firebase for testing. Currently connected test are commented out as they are failing and will require follow up work to fix the native code.
  - iOS CI - upgraded to use the newer and aligned version to the app Xcode
  - Node CI - created this workflow split the node parts from the Android CI, as it didn't make sense to be running here
- Migrated from `react-native-appearance` to the `react-native` DM implementation. As the former lib has been deprecated and recommended to use the native implementation. Wish also reduces dependencies 🎉 
- Updated the iOS Podfile deps to align to the changes provided by RN with path changes and setup
- Other path changes due to changes to `react-native` 
- Snapshots updated due to removal of tvOS being removed from the RN spec


lib/bpk-appearance/src/BpkAppearance.js
```diff
@@ -17,38 +17,30 @@
...
-
-import { initialMode, eventEmitter } from 'react-native-dark-mode';
-
-export type ColorSchemeName = 'light' | 'dark';
+import { Appearance } from 'react-native';
+import type { AppearancePreferences } from 'react-native/Libraries/Utilities/NativeAppearance';
 
 export type BpkAppearancePreferences = {
-  colorScheme?: ColorSchemeName,
+  colorScheme?: AppearancePreferences,
```
olliecurtis, line 24: `AppearancePreferences` type is the exact same as the previous self defined `ColorSchemeName` so utilising the platform type to ensure correctness

storybook/storybook.js
```diff
@@ -56,11 +50,6 @@ const initRtlAddon = (channel) => {
   channel.on(RTL_EVENT, (rtlEnabled) => I18nManager.forceRTL(rtlEnabled));
 };
 
-const initDarkModeAddon = (channel) => {
```
olliecurtis, line 59: Removed this as addons are not even available in the RN storybook.

jest/setup.js
```diff
@@ -18,14 +18,17 @@
...
+import { useState } from 'react';
 import { NativeModules } from 'react-native';
 
-NativeModules.RNDarkMode = {
-  initialMode: 'light',
-  supportsDarkMode: true,
-  addListener: jest.fn(),
-  removeListeners: jest.fn(),
-};
+jest.mock('react', () => ({
```
olliecurtis, line 24: We mock hooks here, as it helps with mocking of the RN `Appearance` module that is using hooks

lib/android/build.gradle
```diff
@@ -67,11 +67,4 @@ dependencies {
   implementation "com.github.skyscanner:backpack-android:${_backpackAndroidVersion}"
   implementation "androidx.constraintlayout:constraintlayout:1.1.3"
   implementation "com.jakewharton.threetenabp:threetenabp:${rootProject.ext.threetenabpVersion}"
-
-  if (_internalBuild) {
```
olliecurtis, line 71: As we are using the `react-native` dark mode implementation we no longer need to bundle the third party dep as its no longer used in the app 🎉

lib/android/src/androidTest/java/net/skyscanner/backpack/reactnative/rating/RNBpkRatingTest.kt
```diff
@@ -54,7 +54,7 @@ class RNBpkRatingTest {
     rating.state.icon = icon
     rating.render()
 
-    verify { uiModule.setViewLocalData(eq(rating.id), any()) }
+    verify { uiModule?.setViewLocalData(eq(rating.id), any()) }
```
olliecurtis, line 57: An error was being thrown that this could be null and would cause tests to fail

lib/android/src/main/java/net/skyscanner/backpack/reactnative/calendar/CalendarViewManager.kt
```diff
@@ -136,7 +136,7 @@ class CalendarViewManager : ViewGroupManager<RNCalendarView>() {
 
   @VisibleForTesting
   public override fun addEventEmitters(reactContext: ThemedReactContext, view: RNCalendarView) {
-    val dispatcher = reactContext.getNativeModule(UIManagerModule::class.java).eventDispatcher
+    val dispatcher = reactContext.getNativeModule(UIManagerModule::class.java)?.eventDispatcher
```
olliecurtis, line 139: An error was being thrown that this could be null and would cause tests to fail

lib/android/src/main/java/net/skyscanner/backpack/reactnative/calendar/CalendarViewManager.kt
```diff
@@ -150,7 +150,7 @@ class CalendarViewManager : ViewGroupManager<RNCalendarView>() {
         }
       }
 
-      dispatcher.dispatchEvent(CalendarChangeEvent(view.id, dates.toTypedArray()))
+      dispatcher?.dispatchEvent(CalendarChangeEvent(view.id, dates.toTypedArray()))
```
olliecurtis, line 153: An error was being thrown that this could be null and would cause tests to fail

lib/android/src/main/java/net/skyscanner/backpack/reactnative/dialog/DialogViewManager.kt
```diff
@@ -103,10 +103,10 @@ class DialogViewManager : ViewGroupManager<RNDialog>() {
   }
 
   public override fun addEventEmitters(reactContext: ThemedReactContext, view: RNDialog) {
-    val dispatcher = reactContext.getNativeModule(UIManagerModule::class.java).eventDispatcher
+    val dispatcher = reactContext.getNativeModule(UIManagerModule::class.java)?.eventDispatcher
```
olliecurtis, line 106: An error was being thrown that this could be null and would cause tests to fail

lib/android/src/main/java/net/skyscanner/backpack/reactnative/dialog/DialogViewManager.kt
```diff
@@ -103,10 +103,10 @@ class DialogViewManager : ViewGroupManager<RNDialog>() {
...
 
     view.state.onAction = { type, pos ->
-      dispatcher.dispatchEvent(DialogActionEvent(view.id, type, pos))
+      dispatcher?.dispatchEvent(DialogActionEvent(view.id, type, pos))
```
olliecurtis, line 109: An error was being thrown that this could be null and would cause tests to fail

lib/android/src/main/java/net/skyscanner/backpack/reactnative/rating/RNBpkRating.kt
```diff
@@ -42,7 +42,7 @@ class RNBpkRating(
     val uiManager = reactContext.getNativeModule(UIManagerModule::class.java)
     val localData = BpkRatingLocalData(view, state.orientation, state.size)
     // This will trigger measure to run in BpkRatingShadowNode
-    uiManager.setViewLocalData(id, localData)
+    uiManager?.setViewLocalData(id, localData)
```
olliecurtis, line 45: An error was being thrown that this could be null and would cause tests to fail

package.json
```diff
@@ -94,25 +94,26 @@
...
+    "jest-expo": "^40.0.1",
     "lint-staged": "^10.5.4",
     "markdown-spellcheck": "^1.3.1",
-    "metro-react-native-babel-preset": "^0.63.0",
+    "metro-react-native-babel-preset": "^0.64.0",
     "prop-types": "^15.5.10",
-    "react": "16.9.0",
-    "react-dom": "16.9.0",
-    "react-native": "0.61.5",
-    "react-native-dark-mode": "^0.2.2",
```
olliecurtis, line 110: Gone 🎉

.github/workflows/ios-ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: select Xcode version
-        run: sudo xcode-select -switch /Applications/Xcode_11.7.app
+        run: sudo xcode-select -switch /Applications/Xcode_12.4.app
```
shaundon, line 18: Should we change this to 12.5 now?

ios/Rakefile
```diff
@@ -1,5 +1,5 @@
 FULL_TESTS = ENV['FULL_TESTS'] != 'false'
-BUILD_SDK = ENV['BUILD_SDK'] || 'iphonesimulator13.7'
+BUILD_SDK = ENV['BUILD_SDK'] || 'iphonesimulator14.4'
```
shaundon, line 2: Same as before – is it worth us making this 14.5?

shaundon:
I'm pretty sure, but just want to be certain – switching the BpkAppearance implementation from a third party to first party is non breaking for users, right? As in they won't need to do anything, we've handled the change behind the scenes

.github/workflows/ios-ci.yml
olliecurtis: Good question! I will need to check as the reason for this was to align with the CI of our Backpack App and also the Skyscanner App which does not yet support Xcode 12.5 so making sure conditions are similar

ios/Rakefile
olliecurtis: This was to align with what we are using for the Backpack iOS app so not sure if it would cause issues if there are mis aligned

olliecurtis:
> I'm pretty sure, but just want to be certain – switching the BpkAppearance implementation from a third party to first party is non breaking for users, right? As in they won't need to do anything, we've handled the change behind the scenes

@shaundon yes that is correct its all the underlying implementation that has changed and nothing on consumer end

shaundon (approved):

.github/workflows/android-ci.yml
```diff
@@ -11,55 +11,49 @@ defaults:
...
+      - name: Gradle Lint
+        run: ./android/gradlew -p android ktlint
+
+      - name: Gradle build
+        run: ./android/gradlew -p android :app:assembleDebug :backpack-react-native:assembleAndroidTest
+
+      - name: Run Android unit tests
+        run: ./android/gradlew -p android :app:Test :backpack-react-native:Test
+  
+  # These tests currently fail so we will need a follow up ticket to fix them.
```
georgegillams, line 39: Is this how we're leaving them? Is there a follow-up ticket we can reference here?

.github/workflows/ios-ci.yml
georgegillams: We are now using xcode 12.5 everywhere 👍

.github/workflows/android-ci.yml
olliecurtis: I need to create that, thanks for the reminder! We never actually used to run these tests like this so was a new addition but need Android support to take a look and understand the error!

olliecurtis: I do however think I might know what the problem is looking at it, so let me take another look first and then if its still an issue will create the appropriate backlog ticket

android/app/build.gradle
```diff
@@ -162,7 +162,7 @@ android {
             def abi = output.getFilter(OutputFile.ABI)
             if (abi != null) {  // null for the universal-debug, universal-release variants
                 output.versionCodeOverride =
-                        versionCodes.get(abi) * 1048576 + defaultConfig.versionCode
+                defaultConfig.versionCode * 1000 + versionCodes.get(abi)
```
georgegillams, line 165: What was 1048576? Can we have a comment explaining what this maths is doing?

olliecurtis: Sure, I will dig it out but this was just a recommendation that came from the react native upgrade guide

ios/Rakefile
georgegillams: When we upgrade to xcode 12.5 then lets also change this to iOS 12.5

ios/native/AppDelegate.m
```diff
@@ -35,7 +35,14 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
                                                  initialProperties:nil
                                                      launchOptions:launchOptions];
     rootView.backgroundColor = [[UIColor alloc] initWithRed:1.0f green:1.0f blue:1.0f alpha:1];
-
+    
+    // Adding here for when we migrate to native RN DM support
```
georgegillams, line 39: Surely we've done that already in this PR?

olliecurtis: Yes good spot need to re instate this now! From an iOS perspective do these colours match what we do on Backpack iOS

android/app/build.gradle
olliecurtis: https://react-native-community.github.io/upgrade-helper/?from=0.61.5&to=0.64.0

This is just something and config from react native that they require I don't know the 100% maths they came up with but it looks like there is a comment above explaining what this section and function is

https://github.com/Skyscanner/backpack-react-native/pull/715/files/05d85662efa8e35f3a589bd8f0713d344ba71f72#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R159

olliecurtis: Screenshot 2021-08-02 at 12 53 08

lib/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": false,
   "name": "backpack-react-native",
-  "version": "14.1.2",
+  "version": "15.0.0-beta.1",
```
georgegillams, line 4: Not sure if this should have changed as it's an unofficial release?

package.json
```diff
@@ -23,14 +23,14 @@
...
-    "flow-check-android": "flow check --flowconfig-name .flowconfig.android",
+    "flow": "npm run flow-check-ios && npm run flow-check-android",
+    "flow-check-ios": "flow check --show-all-errors --show-all-branches",
+    "flow-check-android": "flow check --show-all-errors --show-all-branches --flowconfig-name .flowconfig.android",
     "check-bpk-dependencies": "",
     "fix-bpk-dependencies": "",
     "test": "npm run lint && npm run check-bpk-dependencies && jest --coverage && npm run flow-check-ios && npm run flow-check-android && npm run spellcheck",
     "check-owners": "npm whoami && ( node scripts/npm/get-owners.js | grep -E $(npm whoami) ) && node scripts/npm/check-owners.js",
-    "release": "npm run check-owners && git fetch && git pull --rebase && npm run flow stop && npm test && npm run android:test && npm run build && node scripts/npm/release/index.js",
+    "release": "npm run check-owners && git fetch && git pull --rebase && npm run flow && npm test && npm run android:test && npm run build && node scripts/npm/release/index.js",
```
georgegillams, line 33: `npm run flow stop` was here to ensure that we don't have duplicated servers running locally - as it can break things.
`npm test` runs `flow` so no need to run here

olliecurtis: Good catch, unsure why I had changed that onw!

ios/native/AppDelegate.m
olliecurtis: Wondering if we actually need this at all, given that the app seems to work fine and respect the correct colours already without it?

I had forgotten to address this so wondering if its still relevant

georgegillams (approved):

PR #728, opened 2021-02-11
[BPK-4392]: Migrate dark mode

olliecurtis:
The purpose of this PR is to migrate our dark mode implementation from a 3rd party library to the native solution provided by React Native >0.62.0.

The changes in this should not have an impact on consumers as the changes happen under to hood so the API remains the same.

This PR is being made to #715 on purpose as this Native support is not available in the current version of RN we are using and has come as part of the newer version upgrades that we are undertaking.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests
+ [N/A] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)

### DIFF
--- a/ios/native/Info.plist
+++ b/ios/native/Info.plist
@@ -36,7 +36,7 @@
 		</dict>
 	</dict>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string></string>
+	<string/>
 	<key>UIAppFonts</key>
 	<array>
 		<string>BpkIcon.ttf</string>
--- a/ios/native/Info.plist
+++ b/ios/native/Info.plist
@@ -36,7 +36,7 @@
 		</dict>
 	</dict>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string/>
+	<string></string>
 	<key>UIAppFonts</key>
 	<array>
 		<string>BpkIcon.ttf</string>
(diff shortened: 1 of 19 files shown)


ios/native/Info.plist
```diff
@@ -36,7 +36,7 @@
 		</dict>
 	</dict>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string/>
+	<string></string>
```
georgegillams, line 39: This change seems pointless

olliecurtis: This looks like it is something automatic that xcode did or formatting did. I didn't touch it but will try change it back!

olliecurtis: Solved!

lib/bpk-appearance/src/BpkAppearance-test.js
```diff
@@ -15,98 +15,82 @@
...
   });
 
   it('correctly returns initial appearance', () => {
-    expect(subject.get()).toEqual({ colorScheme: 'light' });
+    expect(appearance.get()).toEqual({ colorScheme: 'light' });
   });
 
   it('correctly updates the appearance', () => {
-    expect(subject.get()).toEqual({ colorScheme: 'light' });
+    expect(appearance.get()).toEqual({ colorScheme: 'light' });
```
georgegillams, line 32: this test seems the same as the above?

olliecurtis: You are right! Hadn't spotted that, have removed it

georgegillams (approved):

PR #755, opened 2021-03-12
[No Jira] Tweak BpkNavigationBar to support iPhone 12 device family

shaundon:


lib/bpk-component-navigation-bar/src/BpkNavigationBar.ios.js
```diff
@@ -171,7 +170,7 @@ class BpkNavigationBar extends Component<EnhancedProps, {}> {
     const styles = unpackBpkDynamicValue(dynamicStyles, bpkAppearance);
     const titleStyle = [styles.title];
     const outerBarStyle = [styles.barOuter];
-    const innerBarStyle = [styles.barInner, isIphoneX && styles.iPhoneXBar];
+    const innerBarStyle = [styles.barInner];
```
shaundon, line 173: Simplified the styles above to remove the need for this logic. I think this way is a lot more understandable, I was initially quite confused by the logic in the styles.

georgegillams (approved):

PR #759, opened 2021-03-16
[BPK-4433] Calendar accessibility improvements(iOS)

k0nserv:
This PR adopts and aligns RN with the accessibility improvements on iOS. It doesn't change anything about Android. Sibling PR: https://github.com/Skyscanner/backpack-ios/pull/919

My plan is that we'll evolve the api to require consumers to call one of the `make*` functions with the appropriate arguments required by Flow etc.

Usage is would be something like 

```javascript
<BpkCalendar selectionType={makeSingleSelection(I18n.localize('CALENDAR_SINGLE_SELECTION_HINT')} {..rest}
```

I'm mostly interested in a review of this API. For a consumer the change will be switching from something like `SELECTION_TYPES.single` to the above. 

Other options:

* Make each selection type a JS class with the same flow typing. Would require a `new` which doesn't really work nicely inside the middle of a component being rendered. This could possibly mean we don't need the `_type` field as we could use `instanceof`. However, we still need something like `_type` to distinguish between the different kinds on the native side.

+ [x] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-react-native/blob/main/CONTRIBUTING.md)

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Create a migration guide
+ [x] Write tests for `assertStrings` and other new introductions in `common-types.js`
+ [x] Tests

_If you are curious about how we review, please read through the [code review guidelines](https://github.com/Skyscanner/backpack/blob/main/CODE_REVIEW_GUIDELINES.md)_

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+**Breaking:**
+
+- Improved accessibility of the calendar, `BpkCalendar`. When setting the `selectionType` prop it's no longer enough to use e.g. `SELECTION_TYPES.single`. Instead of `SELECTION_TYPES` there are three new functions to use: `makeSingleSelection`, `makeRangeSelection`, `makeMultipleSelection`. These expect a few additional string arguments for assistive technology users. Read more in the [migration guide](docs/14.0.0-calendar-accessibility-migration.md)
+
+
+
 > Place your changes below this line.
 
 ## How to write a good changelog entry
--- a/docs/14.0.0-calendar-accessibility-migration.md
+++ b/docs/14.0.0-calendar-accessibility-migration.md
@@ -1,0 +1,162 @@
+# 14.0.0 BpkCalendar Migration Guide
+
+This guide explains how to migrate use of `BpkCalendar` in the `14.0.0` release.
+
+In common for all the scenarios described below is the fact that you'll need to provide more strings for accessibility purposes. These strings should be localised in the traveller's language.
+
+
+**Note:** All of the relevant strings are required, this is enforced via prop types and Flow. The component will throw an error if any are missing or empty.
+
+## Single Selection
+
+Single selection is the simplest case.
+
+### Before
+
+```javascript
+import React, { Component } from 'react';
+import BpkCalendar, { SELECTION_TYPES } from 'backpack-react-native/bpk-component-calendar';
+
+const MyCalendar = (props) => <BpkCalendar selectionType={SELECTION_TYPES.single} {...rest} />;
+
+```
+
+### After
+
+```javascript
+import React, { Component } from 'react';
+import BpkCalendar, { makeSingleSelection } from 'backpack-react-native/bpk-component-calendar';
+
+const singleSelection = makeSingleSelection({
+
+  // `CALENDAR_SINGLE_SELECT_HINT_LABEL` should be a string that provides a
+  // hint to the user on how to select a date.
+  //
+  // English example: "Double tap to select date"
+  selectHint: I18n.translate('CALENDAR_SINGLE_SELECT_HINT_LABEL'),
+});
+
+const MyCalendar = (props) => <BpkCalendar selectionType={singleSelection} {...rest} />;
+
+```
+
+## Range Selection
+
+Range selection is very complex and requires quite a few strings to provide a good screen reader experience.
+
+### Before
+
+```javascript
+import React, { Component } from 'react';
+import BpkCalendar, { SELECTION_TYPES } from 'backpack-react-native/bpk-component-calendar';
+
+const MyCalendar = (props) => <BpkCalendar selectionType={SELECTION_TYPES.range} {...rest} />;
+
+```
+
+### After
+
+```javascript
+import React, { Component } from 'react';
+import BpkCalendar, { makeRangeSelection } from 'backpack-react-native/bpk-component-calendar';
+
+const rangeSelection = makeRangeSelection({
+
+  // `CALENDAR_RANGE_SELECT_HINT_LABEL` should be
+  // a hint that is read out to screen reader users
+  // when no dates have been selected. Should explain how
+  // to select the start date i.e. by double tapping.
+  //
+  // English example: "Double tap to select departure date"
+  startDateSelectHint: I18n.translate('CALENDAR_RANGE_START_DATE_SELECT_HINT_LABEL'),
+
+  // `CALENDAR_RANGE_END_DATE_SELECT_HINT_LABEL` should be
+  // a hint that is read out to screen reader users
+  // when only the start date has been selected. Should explain how
+  // to select the end date i.e. by double tapping.
+  //
+  // English example: "Double tap to select return date"
+  endDateSelectHint: I18n.translate('CALENDAR_RANGE_END_DATE_SELECT_HINT_LABEL'),
+
+  // CALENDAR_RANGE_START_DATE_SELECTED_STATE_LABEL should be
+  // a string that is read out to screen reader users
+  // when the selected start date has focus.
+  //
+  // English example: "Selected as departure date"
+  startDateSelectedState: I18n.translate('CALENDAR_RANGE_START_DATE_SELECTED_STATE_LABEL'),
+
+  // CALENDAR_RANGE_END_DATE_SELECTED_STATE_LABEL should be
+  // a string that is read out to screen reader users
+  // when the selected end date has focus.
+  //
+  // English example: "Selected as return date"
+  endDateSelectedState: I18n.translate('CALENDAR_RANGE_END_DATE_SELECTED_STATE_LABEL'),
+
+  // CALENDAR_RANGE_END_AND_START_DATE_SELECTED_STATE_LABEL should be
+  // a string that is read out to screen reader users
+  // when the selected end date is the same as the start date and
+  // this date has focus.
+  //
+  // English example: "Selected as both departure and return date"
+  endAndStartDateSelectedState: I18n.translate('CALENDAR_RANGE_END_AND_START_DATE_SELECTED_STATE_LABEL'),
+
+  // CALENDAR_RANGE_DATE_BETWEEN_START_AND_END_SELECTED_STATE_LABEL
+  // should be a string that is read out to screen reader users
+  // when a date between the start and end date has focus. e.g.
+  // when the start date is 01/03/2021 and the end date is 05/03/2021
+  // and the date with focus is 03/03/2021.
+  //
+  // English example: "Selected between departure and return date"
+  dateBetweenStartAndEndSelectedState: I18n.translate('CALENDAR_RANGE_DATE_BETWEEN_START_AND_END_SELECTED_STATE_LABEL'),
+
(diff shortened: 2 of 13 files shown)


lib/bpk-component-calendar/src/common-types.js
```diff
@@ -116,7 +136,17 @@ export const commonPropTypes = {
...
+    PropTypes.shape({
+      _type: PropTypes.oneOf(['single']),
+    }),
+    PropTypes.shape({
+      _type: PropTypes.oneOf(['range']),
+    }),
+    PropTypes.shape({
+      _type: PropTypes.oneOf(['multiple']),
+    }),
+  ]),
```
georgegillams, line 149: I know this is WIP, but why not use this
```suggestion
  selectionType: PropTypes.shape([
    _type: PropTypes.oneOf(['single', 'range', 'multiple']),
  }),
```

k0nserv: Right now that would work, but when we add the a11y strings these will all be different even though they all have the `_type` field.

georgegillams: Oh it's because each one is gonna have different additional strings.

lib/bpk-component-calendar/src/common-types.js
```diff
@@ -116,7 +217,17 @@ export const commonPropTypes = {
   maxDate: datePropType,
   onChangeSelectedDates: PropTypes.func,
   selectedDates: selectedDatesPropType,
-  selectionType: PropTypes.oneOf(Object.keys(SELECTION_TYPES)),
+  selectionType: PropTypes.oneOfType([
+    PropTypes.shape({
```
k0nserv, line 221: This needs to be updated

lib/bpk-component-calendar/src/common-types.js
```diff
@@ -30,13 +30,120 @@ import { type FooterView } from './footerView';
...
+  // to unselect the date i.e. by double tapping.
+  deselectHint: string,
+|};
+
+export type SelectionType =
+  | SelectionTypeSingle
+  | SelectionTypeRange
+  | SelectionTypeMultiple;
+
+export const assertString = (
```
k0nserv, line 98: I don't know if this is too heavy handed or not. WDYT?

The intent here is to catch any developers who ignore Flow. In a "proper" statically typed language this wouldn't be needed but since Flow errors can simply be ignored it might help

lib/bpk-component-calendar/src/common-types.js
k0nserv: I've updated this now

UNRELEASED.md
```diff
@@ -1,5 +1,11 @@
 # Unreleased
 
+**Breaking:**
+
+- Improved accessibility of the calendar, `BpkCalendar`. When setting the `selectionType` prop it's no longer enough to use e.g. `SELECTION_TYPES.single`. Instead of `SELECTION_TYPES` there are three new functions to use: `makeSingleSelection`, `makeRangeSelection`, `makeMultipleSelection`. These expected a few additional string arguments for screen reader users. Read more in the [migration guide](docs/14.0.0-calendar-accessibility-migration.md)
```
georgegillams, line 5:
```suggestion
- Improved accessibility of the calendar, `BpkCalendar`. When setting the `selectionType` prop it's no longer enough to use e.g. `SELECTION_TYPES.single`. Instead of `SELECTION_TYPES` there are three new functions to use: `makeSingleSelection`, `makeRangeSelection`, `makeMultipleSelection`. These expect a few additional string arguments for assistive technology users. Read more in the [migration guide](docs/14.0.0-calendar-accessibility-migration.md)
```

lib/bpk-component-calendar/src/common-types.js
k0nserv: Done now

ios/Podfile
```diff
@@ -37,7 +37,7 @@ target 'Backpack Native' do
   pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec', :modular_headers => false
 
   # RN Bridging Pods
-  pod 'Backpack', '~> 41.0'
+  pod 'Backpack', git: 'https://github.com/Skyscanner/backpack-ios.git', branch: 'BPK-4426-add-calendar-accessibility-strings'
```
olliecurtis, line 40: Is this so that we can test it works and will be removed once the iOS part is done?

lib/bpk-component-calendar/src/NativeCalendar.ios.js
```diff
@@ -43,6 +43,7 @@ const BpkCalendar = (props: Props) => {
     selectionType,
     ...rest
   } = props;
+  // delete selectionType.selectHint;
```
olliecurtis, line 46: Is this supposed to still be here?

lib/bpk-component-calendar/src/common-types-test.js
```diff
@@ -0,0 +1,131 @@
...
+    it('should return an appropriate object', () => {
+      expect(makeSingleSelection({ selectHint: 'Test' })).toEqual({
+        type: 'single',
+        selectHint: 'Test',
+      });
+    });
+
+    it('should throw when `selectHint` is missing', () => {
+      expect(() => {
+        // $FlowFixMe
```
olliecurtis, line 59: Is this due to the TODOs with the types and finding a better way to do it? Otherwise could you add a comment to say why this has been added here for future reference?

olliecurtis (changes requested):
LGTM - just a couple of minor points, will run it locally to test it now

ios/Podfile
k0nserv: Yeah, I'll update this when we merge and release Skyscanner/backpack-ios#919

lib/bpk-component-calendar/src/NativeCalendar.ios.js
k0nserv: Good catch and no

lib/bpk-component-calendar/src/common-types-test.js
k0nserv: No, it's becaues I am intentionally violating the flow types to cause the throw I want to test. I'll add a comment

docs/14.0.0-calendar-accessibility-migration.md
```diff
@@ -0,0 +1,162 @@
...
+  // hint to the user on how to deselect a selected date.
+  //
+  // English example: "Double tap to deselect date"
+  selectHint: I18n.translate('CALENDAR_SINGLE_DESELECT_HINT_LABEL'),
+});
+
+const MyCalendar = (props) => <BpkCalendar selectionType={multipleSelection} {...rest} />;
+
+```
+
```
shaundon, line 162: No comment other than to say this migration guide is exceptionally well written! Nice work 🎉

lib/bpk-component-calendar/src/common-types.js
```diff
@@ -30,13 +30,148 @@ import { type FooterView } from './footerView';
 type ViewProps = ElementProps<typeof View>;
 type ViewStyleProp = $PropertyType<ViewProps, 'style'>;
 
-export const SELECTION_TYPES = {
-  single: 'single',
-  range: 'range',
-  multiple: 'multiple',
+// NOTE: This arary needs to be kept in sync with
```
shaundon, line 33:
```suggestion
// NOTE: This array needs to be kept in sync with
```

lib/bpk-component-calendar/src/common-types.js
shaundon: I don't think it's too heavy handed. We know that consumers often ignore Flow types, and from an accessibility point of view that is unacceptable here, so we should enforce it. In other words, enforcing it is the most traveller-first thing to do.

lib/bpk-component-calendar/src/common-types-test.js
```diff
@@ -0,0 +1,134 @@
...
+      }).toThrow();
+    });
+  });
+
+  describe('makeMultipleSelection', () => {
+    let expectedArgs;
+
+    beforeEach(() => {
+      expectedArgs = ['selectHint', 'deselectHint'].reduce((acc, k) => {
+        acc[k] = 'Test';
```
georgegillams, line 73: Maybe this should be `Test_${k}`.
At the moment every test value is the same, so if the logic assigns the wrong value then the test would still pass.

lib/bpk-component-calendar/src/common-types-test.js
```diff
@@ -0,0 +1,134 @@
...
+      expectedArgs = [
+        'startDateSelectHint',
+        'endDateSelectHint',
+        'startDateSelectedState',
+        'endDateSelectedState',
+        'endAndStartDateSelectedState',
+        'dateBetweenStartAndEndSelectedState',
+        'makeNextSelectionPrompt',
+      ].reduce((acc, k) => {
+        acc[k] = 'Test';
```
georgegillams, line 108: likewise

georgegillams (changes requested):
Looking great. Just a couple of suggestions from me 👍

lib/bpk-component-calendar/index.js
```diff
@@ -19,7 +19,15 @@
 /* @flow */
 
 import BpkCalendar, { type Props } from './src/BpkCalendar';
-import { SELECTION_TYPES, type SelectionType } from './src/common-types';
+import {
+  makeSingleSelection,
+  makeRangeSelection,
+  makeMultipleSelection,
+  type SelectionType,
```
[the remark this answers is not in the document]
k0nserv: I think it's still useful? For example if a consumer want to write a function like

```javascript
const makeSelectionTypeFromConfig = (config: [string]: any): SelectionType => {}
```

If we don't export SelectionType they'd need to write it as `SelectionTypeSingle | SelectionTypeRange | SelectionTypeMultiple`

lib/bpk-component-calendar/src/common-types.js
```diff
@@ -30,13 +30,148 @@ import { type FooterView } from './footerView';
...
+// SelectionTypeSingle below.
+//
+// TODO: Figure out if there's a way to create the
```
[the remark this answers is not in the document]
k0nserv: I couldn't figure out a way to do it no

lib/bpk-component-calendar/src/common-types-test.js
k0nserv: Good shout

lib/bpk-component-calendar/src/common-types-test.js
k0nserv: likewise

lib/bpk-component-calendar/src/common-types.js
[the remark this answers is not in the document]
georgegillams: Do you wanna clean up the TODO comments then?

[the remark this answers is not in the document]
k0nserv: I feel like it's worth leaving in place if we move to TS and find it's possible in TS or if Flow evolves to support it

olliecurtis (approved):

georgegillams (approved):

PR #773, opened 2021-04-14
[KOA-4486]: Adding new zIndex prop to map markers

olliecurtis:
Adding new `zIndex` prop to fix an issue where when a map marker was selected in a cluster of markers it would be selected behind unselected markers.

| **Price Marker Before** | **Price Marker After** |
| -- | -- |
| Screenshot_20210414-114256 | Screenshot_20210414-114104

| **Icon Marker Before** | **Icon Marker After** |
| -- | -- |
| Screenshot_20210414-114320 | Screenshot_20210414-114133

+ [x] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-react-native/blob/main/CONTRIBUTING.md)

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

_If you are curious about how we review, please read through the [code review guidelines](https://github.com/Skyscanner/backpack/blob/main/CODE_REVIEW_GUIDELINES.md)_

### DIFF
--- a/lib/bpk-component-map/src/BpkIconMarker.android.js
+++ b/lib/bpk-component-map/src/BpkIconMarker.android.js
@@ -39,7 +39,15 @@ import {
 const dynamicStyles = BpkDynamicStyleSheet.create(androidStyles);
 
 const BpkIconMarker = (props: Props) => {
-  const { disabled, icon, latitude, longitude, onPress, status } = props;
+  const {
+    disabled,
+    icon,
+    latitude,
+    longitude,
+    onPress,
+    status,
+    zIndex,
+  } = props;
   const styles = useBpkDynamicStyleSheet(dynamicStyles);
 
   const wrapperStyles = [styles.wrapper];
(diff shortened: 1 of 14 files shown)


lib/bpk-component-map/src/BpkIconMarker.android.js
```diff
@@ -39,7 +39,15 @@ import {
...
 const BpkIconMarker = (props: Props) => {
-  const { disabled, icon, latitude, longitude, onPress, status } = props;
+  const {
+    disabled,
+    icon,
+    latitude,
+    longitude,
+    onPress,
+    status,
+    zIndex,
```
georgegillams, line 49: Why isn't this just spread with `rest`?

olliecurtis: I guess because we called it out as a user prop and goes into a specific part of the view.

Unless you mean more generally with the props?

georgegillams: Yeah I'm wondering why we weren't spreading props before. Seems we shouldn't have had to change the component to allow this.

georgegillams (approved):

PR #789, opened 2021-05-05
Migrating CI methods to align with Backpack Android

olliecurtis:
As the CI keeps failing with the gcloud/ndk setup, here we are updating CI to align closer to the Backpack Android CI.

Currently the connected Android tests fail so for now commenting them out and will create a follow up ticket for the Android experts to help take a look.


.github/workflows/android-ci.yml
```diff
@@ -11,55 +11,50 @@ defaults:
...
-          rm -rf $HOME/google-cloud-sdk
-          export CLOUDSDK_CORE_DISABLE_PROMPTS=1
-          curl https://sdk.cloud.google.com | bash
-        fi
-
-        gcloud version
-        gcloud auth activate-service-account --project $GCLOUD_PROJECT_ID --key-file $HOME/gcloud-service-key.json
-    
-    # We install the NDK as the app requires an older version than bundled in the image
-    - name: Install NDK
```
marianeum, line 37: is this no longer needed?

olliecurtis: Not anymore 😄 seems to be fine without it as the bundled version in the VM has it now so we don't need to install/override it

olliecurtis:
Not sure why this is failing as its being proved working as expected here on this RN upgrade branch 🤔 https://github.com/Skyscanner/backpack-react-native/runs/2511223983?check_suite_focus=true

olliecurtis:
Closing this as RN 64 upgrade has already completed and merged which had these changes in already

PR #822, opened 2021-07-20
[KOA-4655] Upgrade backpack-ios to 44.0.0

georgegillams:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests
+ [x] Any changes have been tested on both Android and iOS.


lib/ios/BackpackReactNative/BackpackReactNative.podspec
```diff
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/Skyscanner/backpack-react-native.git" }
   s.source_files  = "Classes/**/*.{h,m}"
 
-  s.dependency 'React'
+  s.dependency 'React-Core'
```
georgegillams, line 18: I changed this to make the BRN project compatible with `use_frameworks!`

package.json
```diff
@@ -111,7 +111,7 @@
     "react-native-codegen": "0.0.7",
     "react-native-dash": "0.0.11",
     "react-native-dotenv": "^2.5.1",
-    "react-native-linear-gradient": "^2.5.6",
+    "react-native-linear-gradient": "https://github.com/react-native-linear-gradient/react-native-linear-gradient#0e3af58",
```
georgegillams, line 114: Had to reference this commit directly as they've made a fix to be compatible with `use_frameworks!` but haven't released for months!

olliecurtis: If this is the case we will need to do some work in the app, for building this specific version. There is some specific artifacts we build for Android manually and publish that we do against this version

PR #849, opened 2021-09-09
Update contributing guide for Android

marianeum:
Some required changes to the contributing guide after upgrading to the latest backpack-android version

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests
+ [x] Any changes have been tested on both Android and iOS.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,7 +70,7 @@ Once you have Ruby, install [Bundler](https://bundler.io) with `gem install bund
 
 Get [Homebrew](https://brew.sh/) if you don't already have it.
 
-Install Watchman with `brew install watchman`, then install Java 8 with `brew cask install adoptopenjdk/openjdk/adoptopenjdk8`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -86,20 +86,20 @@ source ~/.bash_profile
 Accept the SDK licences:
 
 ```
-$ANDROID_SDK_ROOT/tools/bin/sdkmanager --licenses
+$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --licenses
```
olliecurtis, line 89: Does this mean the paths changed in a newer version of the SDK?

In which case would there be a minimum version of supported SDK we need to mention, as my path is still the old one and it works

marianeum: I completely forgot about this PR 😂 it had to do with the switch to Java 11 - I don't remember the details of what went wrong and where but there were issues with the old sdkmanagers with Java 11

frugoman (approved):

PR #942, opened 2022-04-28
[no jira]: Add arm64 support for Android

olliecurtis:
Adds support for generating and running the Backpack React Native on Android for ARM archs.


olliecurtis:
We did this in #998 so closing

PR #949, opened 2022-05-10
[KOA-4189] fix disabled highlighted dates calendar

gert-janvercauteren:
`RCTBPKDateMatcher` uses isBeforeDate and isAfterDate from `RCTBPKCalendarDateUtils`. 

These functions are 'non-inclusive', we have added a new 'isSameDate' function to make sure the start and end date are included when `range` is used on the iOS calendar

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -4,7 +4,7 @@ PODS:
     - FSCalendar (~> 2.8.2)
     - MBProgressHUD (~> 1.2.0)
     - TTTAttributedLabel (~> 2.0.0)
-  - BackpackReactNative (20.0.2):
+  - BackpackReactNative (20.1.0):
     - Backpack
     - React
     - react-native-maps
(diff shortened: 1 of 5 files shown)


ios/Podfile.lock
```diff
@@ -436,4 +436,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: fcf9d1ebed9e236314299522a2ebb30069ee9556
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.10.1
```
olliecurtis, line 439: Is this expected to have been downgraded?

olliecurtis (approved):
LGTM!

ios/Podfile.lock
gert-janvercauteren: No, that's me running an older version. I will push a fix for this soon:tm:

PR #998, opened 2022-09-26
[KOA-5564] Update contribution guide

anambl:
- Update contribution guide
  - Update Android guidelines to install emulator with API level 29 (the lowest version to support dark mode) to be able to test in dark mode
- Update `cocoapods` version to `1.11.3`
- Add support for `arm64` for Android

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


CONTRIBUTING.md
```diff
@@ -92,16 +94,34 @@ $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --licenses
...
 
 ```
-$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager "system-images;android-28;google_apis;x86"
+# x86
+$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager "system-images;android-29;google_apis;x86"
 $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager "system-images;android-24;google_apis;x86"
+
+# ARM
+$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager "system-images;android-29;google_apis;arm64-v8a"
+$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager "system-images;android-24;google_apis;arm64-v8a"
```
olliecurtis, line 103: Wondering if we need to go as low as Andorid 24 and we should set our minimum to the Skyscanner minimum @marianeum can you confirm the Skyscanner base version or if we should do this?

Whats Backpack Androids settings?

android/app/build.gradle
```diff
@@ -117,7 +117,7 @@ android {
         versionCode 1
         versionName "1.0"
         ndk {
-            abiFilters "armeabi-v7a", "x86"
+            abiFilters "arm64-v8a", "x86"
```
olliecurtis, line 120: I remember when previously running we had needed all three instead of just two of them so something like

`"armeabi-v7a", "x86", "arm64-v8a"` just to keep it compatible, see this PR https://github.com/Skyscanner/backpack-react-native/pull/942

olliecurtis (changes requested):

CONTRIBUTING.md
marianeum: Yep 24 is the current minimum, so that's correct.

android/app/build.gradle
anambl: Ok, added that back now 🙂

marianeum (approved):

olliecurtis (approved):

